Issue #63, opened 2018-08-31
broadcast blocks before add_block


qcdll:
current measurements on loadtest shows 10x+ propagation latency when loadtesting, we now try broadcasting blocks right after POW validation, instead of waiting for local add_block

for slave:
1. move broadcast before add_block
2. allow peers to download blocks that are in a pool, instead of just from the chain

ninjaahhh:
should we close this? @qcdll

qcdll:
need to collect data for root block and do the same thing there if needed

PR #65, opened 2018-09-04
broadcast new blocks before adding to local state

qcdll:
watch local effectiveness:
add ```await asyncio.sleep(0.2)``` to save.add_block(), and check if peer can still download the block


qcdll:
#63

Issue #34, opened 2018-08-24
Add mining functions


qcgg:
Add ethash, sha3sha3 to `cluster/miner.py`.
The specific mining function for each shard can be configured through `ShardConfig` in `config.py`
Block validation function shall be updated accordingly.

ninjaahhh:
two extra endpoints to implement (mining algorithm agnostic)

`getWork`: () -> [header hash, block number, difficulty]
`submitWork`: (header hash, nonce, mix digest) -> bool

ninjaahhh:
close this in favor of #105

PR #85, opened 2018-09-14
[ethash p2] Add mining logic with ethash

ninjaahhh:
test case verified also in go-ethereum. `pyethash` c++ not tested yet.


ninjaahhh:
#34

PR #87, opened 2018-09-14
[ethash p3] Support pyethash

ninjaahhh:
note that will fallback to python implementation during testing, because `pyethash` doesn't allow overriding cache & dataset size.

a side note, `pyethash` is order-of-magnitude faster than the python implementation. I'm considering might as well removing the python implementation. but it provides a nice way to understand how ethash works


ninjaahhh:
#34

PR #91, opened 2018-09-17
[ethash p4] Hook up eth miner and mining in consensus

ninjaahhh:
refactored how mining-specific params are passed.

adding tests now.


ninjaahhh:
#34

PR #99, opened 2018-09-18
Add big uint type and use it for difficulty

### LINKED ISSUE #94: Large difficulty value breaks block serialization
it's 32 bits for root and 64 bit for minor block.
need to settle on the right value size for storing difficulty and make sure the difficulty will not overflow.

ninjaahhh:
fixes #94

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -175,15 +175,12 @@ class BigUintSerializer:
 
     def serialize(self, value, barray):
         value_size_in_byte = (value.bit_length() - 1) // 8 + 1
-        if value_size_in_byte >= 256:
-            raise RuntimeError("value size exceeds limit")
-        barray.extend(value_size_in_byte.to_bytes(1, byteorder="big"))
-        barray.extend(value.to_bytes(value_size_in_byte, byteorder="big"))
-        return barray
+        bs = value.to_bytes(value_size_in_byte, byteorder="big")
+        return self.ser.serialize(bs, barray)
 
     def deserialize(self, bb):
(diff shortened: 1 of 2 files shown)


quarkchain/core.py
```diff
@@ -167,6 +167,25 @@ def deserialize(self, bb):
         return [self.ser.deserialize(bb) for i in range(size)]
 
 
+class BigUintSerializer:
+    """Not infinity, but pretty big: 2^2048-1"""
+
+    def __init__(self):
+        self.ser = PrependedSizeBytesSerializer(1)
```
qcgg: not used

quarkchain/tests/test_core.py
```diff
@@ -108,3 +119,25 @@ def test_optional(self):
...
+
+
+class ABigUint(Serializable):
+    FIELDS = [("value", biguint)]
+
+    def __init__(self, value: int):
+        self.value = value
+
+
+class TestBigUint(unittest.TestCase):
```
qcgg: test boundary values like 2 ** (255 * 8) - 1 and 2 ** (255 * 8) - 1  and 0?

quarkchain/core.py
ninjaahhh: lol yeah I probbaly should simply use it

quarkchain/tests/test_core.py
ninjaahhh: done

quarkchain/core.py
ninjaahhh: done

PR #102, opened 2018-09-19
unify logging; fix color-coding and add correct file name

qcdll:
remove absl, also fix color which was broken by calling Logger before initialization in cluster_config.py
also, correctly prints file name of originating caller

### DIFF
--- a/quarkchain/utils.py
+++ b/quarkchain/utils.py
@@ -1,8 +1,11 @@
 import asyncio
 import ctypes
+import io
+import logging
+import os
+import sys
 import time
 import traceback
-import logging
 
 from eth_utils import keccak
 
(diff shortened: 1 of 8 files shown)


quarkchain/utils.py
```diff
@@ -73,28 +76,104 @@ def assert_true_with_timeout(f, duration=1):
...
 
+_LOGGING_FILE_PREFIX = os.path.join("logging", "__init__.")
+
+
+class QKCLogger(logging.getLoggerClass()):
+    """https://github.com/abseil/abseil-py/blob/master/absl/logging/__init__.py
+    refer to ABSLLogger
+    """
+
+    def findCaller(self, stack_info=False):
```
ninjaahhh: not see it being used. is it used internally by the logging module?

qcdll: yeah, `override`

PR #104, opened 2018-09-19
Add double sha256 mining

ninjaahhh:
...also fixed a serialization bug to not include certain fields when mining


quarkchain/cluster/miner.py
```diff
@@ -232,23 +240,62 @@ def mine_ethash(
...
+                    block, _ = input_q.get(block=True)  # blocking
+                    break  # break inner loop to refresh mining params
+                # check if new block arrives. if yes, discard current progress and restart
+                try:
+                    block, _ = input_q.get_nowait()
+                    break  # break inner loop to refresh mining params
+                except Exception:  # queue empty
+                    pass
+                # update param and keep mining
+                start_nonce += rounds
```
qcdll: shall we refresh the block between mining of small batch?

ninjaahhh: line 284 achieves that: if a new block can be retrieved, will refresh to the new block

ninjaahhh:
should close #34

PR #108, opened 2018-09-21
Change timestamp to milliseconds

ninjaahhh:
just to be future-proof.

all timestamps should be regarded as milliseconds unless specified otherwise (like using a `_sec` suffix for variable naming).

### DIFF
--- a/quarkchain/tests/test_config.py
+++ b/quarkchain/tests/test_config.py
@@ -71,7 +71,7 @@ class TestShardConfig(unittest.TestCase):
         {
             "CONSENSUS_TYPE": "POW_SHA3SHA3",
             "CONSENSUS_CONFIG": {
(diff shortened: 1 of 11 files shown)


quarkchain/tests/test_config.py
```diff
@@ -91,7 +91,7 @@ def testBasic(self):
         {
             "CONSENSUS_TYPE": "POW_SHA3SHA3",
             "CONSENSUS_CONFIG": {
-                "TARGET_BLOCK_TIME": 10
+                "TARGET_BLOCK_TIME": 10000
```
qcdll: TARGET_BLOCK_TIME_MS would be less confusing?

qcgg:
shall we close this?

PR #114, opened 2018-09-23
 Fix contract creation in contracts

### LINKED ISSUE #113: Creating contract during function call failed
the tx will disappear instead of failing.

ninjaahhh:
root cause: when contract is created during contract execution,
`to_full_shard_id` is not specified, thus
`mk_contract_address(msg.sender, msg.to_full_shard_id, nonce)` will fail
since rlp can't encode None.

fixes #113

### DIFF
--- a/quarkchain/evm/vm.py
+++ b/quarkchain/evm/vm.py
@@ -15,13 +15,13 @@ verify_stack_after_op = False
 #  ######################################
 
 sys.setrecursionlimit(10000)
-log_log = get_logger('eth.vm.log')
-log_msg = get_logger('eth.pb.msg')
-log_vm_exit = get_logger('eth.vm.exit')
-log_vm_op = get_logger('eth.vm.op')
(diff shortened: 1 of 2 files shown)


quarkchain/evm/vm.py
```diff
@@ -683,7 +683,16 @@ def vm_execute(ext, msg, code):
...
                     ingas = all_but_1n(ingas, opcodes.CALL_CHILD_LIMIT_DENOM)
-                create_msg = Message(msg.to, b"", value, ingas, cd, msg.depth + 1)
+                create_msg = Message(
+                    msg.to,
+                    b"",
+                    value,
+                    ingas,
+                    cd,
+                    msg.depth + 1,
+                    from_full_shard_id=msg.from_full_shard_id,
```
qcgg: from_full_shard_id is not used in create_contract.

it should also be msg.to_full_shard_id which is the shard id of the originating contract.

ninjaahhh: done

Issue #105, opened 2018-09-20
Support offload mining through JSONRPC


ninjaahhh:
something similar to `eth_submitWork` and `eth_getWork`, but should also support specifying different mining algorithms which could be shard-specific.

ninjaahhh:
two extra endpoints to implement (mining algorithm agnostic)

getWork: () -> [header hash, block number, difficulty]
submitWork: (header hash, nonce, mix digest) -> bool

PR #122, opened 2018-09-26
[remote mining p1] Scaffold remote mining

ninjaahhh:
update:
1. no inner loop for work generation, purely triggered by mining clients
1. added tests

related to #105

### DIFF
--- a/quarkchain/cluster/miner.py
+++ b/quarkchain/cluster/miner.py
@@ -48,9 +48,15 @@ class MiningAlgorithm(ABC):
 
     def post_process_mined_block(self, block: Union[MinorBlock, RootBlock]):
         """Post-process block to track block propagation latency"""
-        tracking_data = json.loads(block.tracking_data.decode("utf-8"))
-        tracking_data["mined"] = time_ms()
-        block.tracking_data = json.dumps(tracking_data).encode("utf-8")
+        if isinstance(block, RootBlock):
+            extra_data = json.loads(block.header.extra_data.decode("utf-8"))
+            extra_data["mined"] = time_ms()
+            block.header.extra_data = json.dumps(extra_data).encode("utf-8")
+        else:
+            extra_data = json.loads(block.meta.extra_data.decode("utf-8"))
+            extra_data["mined"] = time_ms()
+            block.meta.extra_data = json.dumps(extra_data).encode("utf-8")
+            block.header.hash_meta = block.meta.get_hash()
 
     @staticmethod
     def _log_status(block: Union[RootBlock, MinorBlock]):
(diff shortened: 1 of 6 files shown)


quarkchain/cluster/miner.py
```diff
@@ -227,51 +238,85 @@ def _mine_new_block_async(self):
...
+                    block,
+                    instance.input_q,
+                    instance.output_q,
+                    mining_params,
+                ),
             )
             instance.process.start()
             await handle_mined_block(instance)
 
+        async def loop_for_work_gen(instance: Miner):
```
qcgg: Why not create work on demand i.e. only when get_work is called?
The work can still be cached and only create new one if the latest one was created more than 5 seconds ago.
That way the logic can be driven by external miner rather than a separate `enabled` variable. If the external miner stopped asking for work it will stop creating new work. The current code has to be stopped by an explicitly `setMining=False` JRPC call.

I also feel these code doesn't necessarily belong to the Miner class because it's actually handling the server side logic for miner. If we separate them out we can drop the `only_remote` check and `enabled` check if we change it to demand.

ninjaahhh: makes sense. the current logic sorta mimics go-ethereum, where the `current_work` is generated either when a block is added or every 5 sec, using an event subscription model.

purely driven by clients + caching should be fine. also should find a way to clear current work once a new block has been added, so clients won't receive outdated work.

will refactor the miner server logic to another file, like `remote_miner` or `miner_client`

ninjaahhh: having a second thought: from master server and shard's point of view, whether the miner is actually a mining doesn't really matter, they merely provide two callbacks (`add` and `create`) to the miner class and kick off the mining process (and only pass `get_work` / `submit_work` calls to them if mining is done outside). I'll keep the current `Miner` class, but remove the update-work-every-5-second logic.

quarkchain/cluster/miner.py
```diff
@@ -227,51 +225,87 @@ def _mine_new_block_async(self):
...
+        if not self.remote:
+            raise ValueError("Should only be used for remote miner")
+
+        if header_hash not in self.work_map:
+            return False
+        block = self.work_map[header_hash]
+        header = copy.copy(block.header)
+        header.nonce, header.mixhash = nonce, mixhash
+        try:
+            validate_seal(header, self.consensus_type)
```
qcgg: validate_seal should be put in add_block_async_func?

ninjaahhh: yes it's already been called when validating root block and minor block (`RootState.validate_block_header` and `ShardState.__validate_block` respectively). an additional validation is added here to make sure it's not spam

ninjaahhh: ...and `add_block_async_func` calls those validation functions

PR #124, opened 2018-09-26
Pull extra data field out from meta to header

ninjaahhh:
...so we can include meta hash for mining but not extra_data


ninjaahhh:
agree on the millisecond PR. will not merge ƒor nøw

PR #126, opened 2018-09-26
Pull extra data field out from meta to header

ninjaahhh:
...so we can include meta hash for mining but not extra_data

this will break current testnet, will merge later


qcdll:
going to be conflict after #127 is merged :-P

ninjaahhh:
note we need to wipe out the testnet after merging this PR

PR #128, opened 2018-09-27
move propagation latency tracking out of extra_data to tracking_data

qcdll:
same as #127 
we can still have #126 after this lands


ninjaahhh:
conflicts should only be some name changing

qcdll:
pretty sure tracking data is in block body for both

ninjaahhh:
image

ah you are right, I'm misled by this

PR #133, opened 2018-10-01
Fix difficulty data type in EcoInfo RPC

ninjaahhh:
theoretically non-breaking, but could cause incompatibility problems during deploying if some hosts use the old data types and some use new


ninjaahhh:
safe to deploy, since it's intra-cluster RPC invocation

PR #134, opened 2018-10-01
Refresh root block during mining simulation

### LINKED ISSUE #107: Refresh mining root block when new minor block header comes in
during simulated mining in testnet, whenever the block gets refreshed (~~a new tx for minor block, or~~ a new minor block header for a root block comes in), should pass the new block into `miner.input_q`, such that miner will discard current progress and start mining new block to maximize testnet's performance.

in production, since new block is triggered by mining (`miner.get_work` will generate blocks to mine) which has a 5-second period during each block generation, this is not an issue.

(it's less an issue for minor block since the mining interval is short.)

ninjaahhh:
another way to fix #107: instead of refreshing root block, simply move logic in #132 to the simulation miner, so other types of miners can work as expected.

should merge this PR then #132

### DIFF
--- a/quarkchain/cluster/miner.py
+++ b/quarkchain/cluster/miner.py
@@ -153,6 +153,8 @@ class DoubleSHA256(MiningAlgorithm):
         return False
 
     def post_process_mined_block(self, block: Block):
+        if not self.nonce_found:
+            raise RuntimeError("cannot post process since no nonce found")
         block.header.nonce = int.from_bytes(self.nonce_found, byteorder="big")
         super().post_process_mined_block(block)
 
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/miner.py
```diff
@@ -215,6 +219,10 @@ def _mine_new_block_async(self):
                 # start mining before processing and propagating mined block
                 instance._mine_new_block_async()
                 try:
+                    # Root block should include latest minor block headers while it's being mined
+                    # This is a hack to get the latest minor block included since testnet does not check difficulty
+                    if instance.consensus_type == ConsensusType.POW_SIMULATE:
+                        block = await self.create_block_async_func()
```
qcdll: add `post_process_mined_block` here per discussion

ninjaahhh: will do

PR #136, opened 2018-10-03
add cancel token for use in p2p service

qcdll:
learn more about cancel_token in https://github.com/ethereum/asyncio-cancel-token
read more why py-evm needed this: https://vorpus.org/blog/timeouts-and-cancellation-for-humans/
it applies to our project as well


ninjaahhh:
add a README and then link to that repo? can also note some changes we intend to add to the original code

PR #137, opened 2018-10-04
[external miner p1] Refactor mining loop

ninjaahhh:
...required for writing external miners reusing current
`Miner._mine_loop` logic by decoupling block and mining work.


quarkchain/cluster/miner.py
```diff
@@ -263,6 +233,8 @@ def _mine_new_block_async(self):
         # 5 sec interval magic number
         if not self.current_work or now - self.current_work.header.create_time > 5:
             block = await self.create_block_async_func()
+            if not block:
+                raise RuntimeError("Created empty block")
```
qcgg: the message is confusing. shouldn't it be "Failed to create block"?

ninjaahhh:
yes we can have both the local & remote mining logic merged, which requires some changes to correctly channel the information (eg stop mining locally if a valid work is submitted, or clear remote work after mining a local block)

will probably do this in a later PR

quarkchain/cluster/miner.py
ninjaahhh: will do

Issue #115, opened 2018-09-24
Restrict root chain mining


qcgg:
Root block signed by a preconfigured private key shall have a weaker requirement on difficulty. The matching public key is hardcoded and used to validate the signature.

PR #139, opened 2018-10-05
Add signature field to RootBlockHeader

qcgg:
#115

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -11,6 +11,7 @@ from typing import List
 
 import ecdsa
 import rlp
+import eth_keys
 from eth_keys import KeyAPI
 
 import quarkchain.evm.messages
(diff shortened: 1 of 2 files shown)


quarkchain/config.py
```diff
@@ -220,6 +220,7 @@ class QuarkChainConfig(BaseConfig):
 
     PROOF_OF_PROGRESS_BLOCKS = 1
     TESTNET_MASTER_ADDRESS = "199bcc2ebf71a851e388bd926595376a49bdaa329c6485f3"
+    GUARDIAN_PUBLIC_KEY = "00" * 64  # TODO: update this
```
qcdll: huh?

qcgg: we'll put a real key here later to turn on guardian

ninjaahhh:
discussed offline, will need to change `Miner.get_work` to add a parameter specifying whether to get work as guardian (lowered difficulty)

PR #140, opened 2018-10-05
Clean up diff / pow check

ninjaahhh:


quarkchain/cluster/master.py
```diff
@@ -162,12 +161,14 @@ def __has_block_hash(self, block_hash):
...
         for i in range(len(block_header_list) - 1):
-            block, prev = block_header_list[i : i + 2]
-            if block.height != prev.height + 1:
+            header, prev = block_header_list[i : i + 2]
+            if header.height != prev.height + 1:
                 return False
-            if block.hash_prev_block != prev.get_hash():
+            if header.hash_prev_block != prev.get_hash():
                 return False
+            validate_seal(header, consensus_type)
```
qcgg: do the same in shard.py?

qcgg: I'll take care of it.

PR #144, opened 2018-10-05
Validate seal in Shard SyncTask and handle_new_block

qcgg:


quarkchain/cluster/shard.py
```diff
@@ -299,13 +299,17 @@ def __has_block_hash(self, block_hash):
...
+            if header.height != prev.height + 1:
                 return False
-            if block.hash_prev_minor_block != prev.get_hash():
+            if header.hash_prev_minor_block != prev.get_hash():
                 return False
+            shard_id = header.branch.get_shard_id()
+            consensus_type = self.shard.env.quark_chain_config.SHARD_LIST[
+                shard_id
+            ].CONSENSUS_TYPE
+            validate_seal(header, consensus_type)
```
ninjaahhh: do you want to raise exception here or catch it and return False?

ninjaahhh: similar to `master.__validate_block_headers`. whichever way we choose, let's keep them the same

PR #146, opened 2018-10-05
add qkchash

qizhou:
The draft code for qkchash.

### DIFF
--- a/qkchash/make.sh
+++ b/qkchash/make.sh
@@ -1,0 +1,1 @@
+g++ qkchash.cpp -std=gnu++17 -O3
--- a/qkchash/qkchash.cpp
+++ b/qkchash/qkchash.cpp
@@ -1,0 +1,149 @@
+#include <cstdint>
+#include <climits>
+
+#include <array>
+#include <chrono>
+#include <iostream>
+#include <random>
+#include <set>
+
+#include <ext/pb_ds/assoc_container.hpp>
+#include <ext/pb_ds/tree_policy.hpp>
+
+using namespace std;
+using namespace __gnu_pbds;
+
+typedef
+tree<
+    uint64_t,
+    null_type,
+    less<uint64_t>,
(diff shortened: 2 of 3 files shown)


ninjaahhh:
does the verification require having the whole datablock? in ethash, `datablock[p]` can be calculated on the fly using the small 16MB cache instead of 1GB dataset, so verification only has a small memory footprint; in qkc hash, seems even verification will also need the whole large dataset?

qcdll:
it looks the "cache" (oset) used by qkchash is only 1024*64
https://github.com/QuarkChain/pyquarkchain/pull/146/files#diff-dab3d2d99cf439d1b6a9dd24f9e4de0bR32
this is much smaller than the dataset used by ethash so mem footprint is tiny

but this raises questions: shall we make it larger? is there going to be ASIC that is very efficient in dealing with 64K ordered numbers?

qizhou:
Good questions.  There are a couple of thoughts:
1, Since every hash calculation will modify datablock, this means if there are multiple hashes are calculated concurrently, the memory requirement will be N * len(datablock) where N is # of concurrent hashes.  If the full dataset is 1GB, then we need N * 1GB (which should be still fine to miner), and the node requires to have extra 1GB to verify.

2, However, using 1GB datablock, then the algorithm is essentially close to ethash with additional memory write operations.

3, My key goal here is that even with a small datablock (e.g., 64 * 1024 * 8, or even smaller), the performance gain of ASIC over CPU should be limited because such pipeline is difficult to be implemented.  Especially, I want the datablock to be smaller enough to be fit into the cache of CPU, i.e., we want to compare:
- CPU + Cache performance
- ASIC (+ memory if the data could be fitted into registers)

The implementation of FPGA suggests that insert/deletion operations are not friendly to pipeline, however, not sure the latest ASIC implementation could improve that.  I am asking some hardware people to further confirm.

PR #150, opened 2018-10-10
Limit the number of minor blocks in root block

qcgg:
#78


qcdll:
i thought this was done a while ago

Issue #155, opened 2018-10-12
estimate_gas failure should not close connection with master


qcgg:
```
MASTER: I1012 04:33:34.078363 jsonrpc.py:408] {"jsonrpc":"2.0","method":"estimateGas","id":"1","params":{"data":"0xa6f2ae3a","to":"0x837f9A0e7185141D75e37D55e213efd725e76710B10d0E9F","value":"0xde0b6b3a7640000","from":"0xb0997B1309C61D50495b79b2B82D377d862Db367b0c6797f"}}
SLAVE_S31: ESC[1;31mE1012ESC[1;0m 04:33:34.135198 protocol.py:167] Traceback (most recent call last):
SLAVE_S31: File "/code/pyquarkchain/quarkchain/protocol.py", line 165, in __internal_handle_metadata_and_raw_data
SLAVE_S31: await self.handle_metadata_and_raw_data(metadata, raw_data)
SLAVE_S31: File "/code/pyquarkchain/quarkchain/cluster/protocol.py", line 80, in handle_metadata_and_raw_data
SLAVE_S31: await super().handle_metadata_and_raw_data(metadata, raw_data)
SLAVE_S31: File "/code/pyquarkchain/quarkchain/protocol.py", line 152, in handle_metadata_and_raw_data
SLAVE_S31: await self.__handle_rpc_request(op, cmd, rpc_id, metadata)
SLAVE_S31: File "/code/pyquarkchain/quarkchain/protocol.py", line 127, in __handle_rpc_request
SLAVE_S31: resp = await handler(self, request)
SLAVE_S31: File "slave.py", line 430, in handle_estimate_gas
SLAVE_S31: res = self.slave_server.estimate_gas(req.tx, req.from_address)
SLAVE_S31: File "slave.py", line 1136, in estimate_gas
SLAVE_S31: return shard.state.estimate_gas(tx, from_address)
SLAVE_S31: File "/code/pyquarkchain/quarkchain/cluster/shard_state.py", line 1290, in estimate_gas
SLAVE_S31: if hi == cap and not run_tx(hi):
SLAVE_S31: File "/code/pyquarkchain/quarkchain/cluster/shard_state.py", line 1277, in run_tx
SLAVE_S31: evm_tx = self.__validate_tx(tx, evm_state, from_address, gas=gas)
SLAVE_S31: File "/code/pyquarkchain/quarkchain/cluster/shard_state.py", line 255, in __validate_tx
[...]

PR #156, opened 2018-10-12
Fixing estimate_gas

qcgg:
fixing #155
1. estimate_gas should never throw
2. estimate_gas should always return an integer even when
the returned gas limit is not enough for the tx
according to https://github.com/ethereum/wiki/wiki/JSON-RPC#parameters-25

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -1368,33 +1368,32 @@ class ShardState:
             Logger.error_exception()
             return None
 
-    def estimate_gas(self, tx: Transaction, from_address) -> Optional[int]:
+    def estimate_gas(self, tx: Transaction, from_address) -> int:
         """Estimate a tx's gas usage by binary searching."""
         evm_tx_start_gas = tx.code.get_evm_transaction().startgas
         # binary search. similar as in go-ethereum
         lo = 21000 - 1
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -1368,33 +1368,32 @@ def get_logs(
...
-                if run_tx(mid):
-                    hi = mid
-                else:
-                    lo = mid
-            except Exception:
+            if run_tx(mid):
+                hi = mid
+            else:
                 lo = mid
-        if hi == cap and not run_tx(hi):
```
ninjaahhh: this is needed otherwise `run(hi)` won't be tried, if the client specifies the cap.

ninjaahhh: ah I see. will return `hi` anyway

qcgg: we don't need to check the result of `run_tx(hi)`. we have to return `hi` anyway.

ninjaahhh: a second thought, this may not be the best way to handle, say if the client specifies a wrong gas cap (like 21001), the binary search will obviously fail, however the new code will return 21001. I think a better way to handle is return error.

also in the link you provided I don't see returning estimate that will not be enough, it's the other way around, that the estimate could be more than actually needed

PR #159, opened 2018-10-13
add discovery and test from py-evm

qcdll:
this is the final part of the port of py-evm p2p module
adds discovery service that uses udp to form a Kademlia-like network and find peers
to learn more: https://github.com/fjl/p2p-drafts

### DIFF
--- a/quarkchain/p2p/discovery.py
+++ b/quarkchain/p2p/discovery.py
@@ -1,0 +1,1303 @@
+"""
+The Node Discovery protocol provides a way to find RLPx nodes that can be connected to. It uses a
+Kademlia-like protocol to maintain a distributed database of the IDs and endpoints of all
+listening nodes.
+
+More information at https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-discovery
+"""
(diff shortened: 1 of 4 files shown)


qcdll:
note this particular test is causing `Abort trap: 6`:
```python -m pytest quarkchain/p2p -k test_topic_query```

File: quarkchain/p2p/discovery.py [hunk 1]
```diff
@@ -755,7 +755,7 @@ def send_topic_nodes(
             n.address.to_endpoint() + [n.pubkey.to_bytes()]
             for n in nodes)
         max_neighbours = self._get_max_neighbours_per_packet()
-        for batch in cytoolz.partition_all(max_neighbours, encoded_nodes):
+        for batch in toolz.partition_all(max_neighbours, encoded_nodes):
```
qcdll: cytoolz here caused segfault:
```
Fatal error in cpyext, CPython compatibility layer: converting a PyTupleObject into a W_TupleObject, but found NULLs as items
Abort trap: 6
```

PR #161, opened 2018-10-15
stake trie

jyouyj:
The draft code for stake trie --- I start from a random function that takes the previous block hash as the random seed and output an index I, which is larger than zero and smaller than the total number of tokens. Then, I try to rewrite the trie such that this stake trie can return the owner's address efficiently.


qcgg:
can you separate your change from the original commit of the trie so that it's easier to see what you have changed over the original trie code?

qcgg:
just do two commits
1. commit the original trie code
2. commit your change

jyouyj:
I have separated the change by doing two commits. Please check it.

qcgg:
can you describe the purpose of the stake trie and the difference from trie?
That'd help us review the code. thanks!

jyouyj:
A stake trie is a variant of the original trie that supports two addition operations beyond insertion, lookup, and deletion:
get_total_stake(): return the total number of stakes stored in the trie
select_staker(i): find the owner address of ith stake stored in the trie
where i is the random index, which is larger than zero and smaller than the total number of stakes.
This stake trie is designed for our proof of stake consensus (Photon). By using this stake trie, we 
can keep track of the total number of stakes and return the ith owner's address efficiently.

For the implementation, I add an extract variable for each trie node to store the total number of stakes rooted at the node (i.e., the number of tokens below it). More specifically, blank or stake trie node in form of [key, [value, stake]] or [[v0, stake],[v1, stake]..[v15, stake],[v, stake]]. All the above operations that modify the trie must adjust the stake information. Besides, I deleted the useless operations split() and merge() in the original trie.

For the test part, I start from a random function that takes the previous block hash as the random seed and output an index I, which is larger than zero and smaller than the total number of tokens. 
Then, I compare the results from the select_staker() operation and nominal values in the sorted list of elements of the trie. The corresponding test dataset is enclosed in the file fixtures/TrieTests/staketrietest.json.

PR #162, opened 2018-10-16
fix test_monitoring.py; py.test can handle whole quarkchain dir

qcdll:
now run tests with:
```
python -m pytest quarkchain
```
get coverage data with:
```
python -m pytest --cov quarkchain quarkchain
```
pytest is so much better than -m unittest


ninjaahhh:
`-k` is super useful.

PR #170, opened 2018-10-18
Apply guardian for pow check

ninjaahhh:


ninjaahhh:
realized this is not needed, since `block_header.difficulty` should already be adjusted by the miners.

ninjaahhh:
not really...this is needed and also need to change how difficulty is checked.

PR #173, opened 2018-10-23
Fix ethash

ninjaahhh:
discovered during locally running mining.

### DIFF
--- a/ethereum/pow/ethpow.py
+++ b/ethereum/pow/ethpow.py
@@ -2,15 +2,10 @@ import warnings
 from functools import lru_cache
 from typing import Tuple, Optional, List, Union
 
-from eth_utils import big_endian_to_int, int_to_big_endian
+from eth_utils import big_endian_to_int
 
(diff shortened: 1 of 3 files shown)


ethereum/pow/ethpow.py
```diff
@@ -139,7 +140,7 @@ def mine(
 
     cache = cache_gen(cache_size, block_number)
     nonce = start_nonce
-    target = zpad(int_to_big_endian(2 ** 256 // (difficulty or 1) - 1), 32)
+    target = (2 ** 256 // (difficulty or 1)).to_bytes(32, byteorder="big")
```
ninjaahhh: this is the difference, previously it won't guarantee 32 bits. another fix is keep `target` as int, and parse `o[b'result']` at line 148 to int every time for comparison.

also added a test to prevent previous error

ninjaahhh: seems ethash utils assume little endian

PR #179, opened 2018-10-28
[rewards p2] Use tax rate when applying tx

ninjaahhh:
plus some naming / coding style changes

two relevant parts:

1. when applying transactions, fees are multiplied with the reward tax rate (e.g. 0.5)
2. when collecting x-shard tx fees, multiply with the same tax rate

comparing with previous approach (processing fee as usual, then deduct half from coinbase address when finalizing the block), this should reflect gas dynamics more accurately

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import time
 from collections import defaultdict
+from fractions import Fraction
 from typing import Optional, Tuple, List, Union, Dict
 
 from quarkchain.cluster.filter import Filter
@@ -9,7 +10,6 @@ from quarkchain.cluster.miner import validate_seal
 from quarkchain.cluster.neighbor import is_neighbor
 from quarkchain.cluster.rpc import ShardStats, TransactionDetail
 from quarkchain.cluster.shard_db_operator import ShardDbOperator
-from quarkchain.config import NetworkId
 from quarkchain.core import (
     calculate_merkle_root,
     Address,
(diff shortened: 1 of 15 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -556,10 +558,6 @@ def run_block(
                 )
                 raise e
 
-        # Put only half of block fee to coinbase address
-        check(evm_state.get_balance(evm_state.block_coinbase) >= evm_state.block_fee)
-        evm_state.delta_balance(evm_state.block_coinbase, -evm_state.block_fee // 2)
```
ninjaahhh: moved this part of logic to 1) applying tx and 2) collecting x-shard fee, respectively

quarkchain/cluster/tests/test_shard_state.py
```diff
@@ -1181,8 +1182,11 @@ def test_add_block_receipt_root_not_match(self):
         # Should succeed
         state.finalize_and_add_block(b1)
         b1.finalize(evm_state=state.run_block(b1))
-        b1.meta.hash_evm_receipt_root = b"00" * 32
-        self.assertRaises(ValueError, state.add_block(b1))
```
ninjaahhh: this won't assert raise, need to be used as context manager

quarkchain/cluster/tests/test_utils.py
```diff
@@ -61,6 +61,11 @@ def get_test_env(
     env.quark_chain_config.SKIP_ROOT_DIFFICULTY_CHECK = True
     env.cluster_config.ENABLE_TRANSACTION_HISTORY = True
     env.cluster_config.DB_PATH_ROOT = ""
+    # prent breaking previous tests after tweaking default rewards
```
qcgg: # prevent?

quarkchain/config.py
```diff
@@ -191,7 +191,7 @@ class RootConfig(BaseConfig):
     GENESIS = None  # type: RootGenesis
 
     COINBASE_ADDRESS = bytes(24).hex()
-    COINBASE_AMOUNT = 5
+    COINBASE_AMOUNT = 120 * QUARKSH_TO_JIAOZI
```
qcgg: shall we calculate this based on other parameters?

PR #182, opened 2018-10-30
[rewards p3] Collect rewards of root block

ninjaahhh:


ninjaahhh:
@gg I also see `__get_cross_shard_tx_list_by_root_block_hash` adding root block coinbase as a x-shard deposit. in this PR the root block coinbase is added during `ShardState.add_root_block`, which do you think is better?

quarkchain/cluster/root_state.py
```diff
@@ -202,11 +202,15 @@ def create_block_to_mine(self, m_header_list, address, create_time=None):
...
+        coinbase_amount = self.env.quark_chain_config.ROOT.COINBASE_AMOUNT
+        reward_tax_rate = self.env.quark_chain_config.REWARD_TAX_RATE
+        minor_block_fee = 0
         for header in m_header_list:
-            coinbase_amount += header.coinbase_amount
-
-        coinbase_amount = coinbase_amount // 2
+            minor_block_fee += header.coinbase_amount
+        # note the minor block fee is after tax
+        coinbase_amount += round(
```
qcgg: fix precision issue

qcgg:
> @gg I also see `__get_cross_shard_tx_list_by_root_block_hash` adding root block coinbase as a x-shard deposit. in this PR the root block coinbase is added during `ShardState.add_root_block`, which do you think is better?

Updating the evm state in add_root_block will break consensus because different clusters might add this root block at different time and won't have a consistent global view.

PR #184, opened 2018-11-12
add sorted list + binary search insert/delete version of qkchash

qizhou:
Using a sorted vector to find ith smallest element
Using binary search to insert an element to keep the vector's elements ordered

Performance: (for 64K entries, 1000 hashes)
DST: 10s
Sorted vector: 20s

The sorted vector code will be used in python (for verification purpose)

### DIFF
--- a/qkchash/qkchash.cpp
+++ b/qkchash/qkchash.cpp
@@ -1,5 +1,6 @@
-#include <cstdint>
 #include <climits>
+#include <cstdint>
+#include <cstring>
 
 #include <array>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


qkchash/qkchash.cpp
```diff
@@ -94,11 +95,49 @@ void qkc_hash(
...
+        uint64_t p = fnv64(i ^ seed[0], mix[i % mix.size()]);
+        for (uint32_t j = 0; j < mix.size(); j++) {
+            // Find the pth element and remove it
+            uint32_t idx = p % slist.size();
+            new_data[j] = slist[idx];
+            slist.erase(slist.begin() + idx);
+
+            // Generate random data and insert it
+            p = fnv64(p, new_data[j]);
+            slist.insert(std::upper_bound(slist.begin(), slist.end(), p), p);
```
qcgg: make sure we don't add duplicated element.

qizhou: Thanks for pointing this out.   Fixed in the new code.  Please take a look.

qcdll: I don't see the downside of keeping duplicates (except it would be different from the ordered set implementation)

qizhou: The downside is that for ordered set implementation, we have to keep the reference count of a key if a key is inserted twice or more.  In addition, implementing an ordered set with duplicated keys is more complicated.

PR #185, opened 2018-11-12
add qkchash.py

qizhou:
Add python version of qkchash

Performance:
Single verification takes about 0.1 s on my Mac using cpython3

### DIFF
--- a/qkchash/qkchash.py
+++ b/qkchash/qkchash.py
@@ -1,0 +1,168 @@
+import bisect
+import sha3
+import time
+import unittest
+
+FNV_PRIME_64 = 0x100000001b3
+UINT64_MAX = 2 ** 64
+
+CACHE_ENTRIES = 1024 * 64
+ACCESS_ROUND = 128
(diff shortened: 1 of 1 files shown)


qkchash/qkchash.py
```diff
@@ -0,0 +1,168 @@
...
+        for j in range(len(mix)):
+            # Find the pth element and remove it
+            remove_idx = p % len(lcache)
+            new_data.append(lcache[remove_idx])
+            lcache_set.remove(lcache[remove_idx])
+            del lcache[remove_idx]
+
+            # Generate random data and insert it
+            p = fnv64(p, new_data[j])
+            if p not in lcache_set:
```
qcgg: umm, we don't have this in the cpp version?

qcgg: this is correct. we need to fix the cpp version.

qcgg:
will add test to make sure python and cpp versions produce the same results.

PR #209, opened 2018-11-21
add qkchash llrb version

qizhou:
Implemented a customized version of left-leaning red-black tree for qkchash with the following improvements:
- The nodes are allocated/freed in a contiguous memory space with freelist.  This makes node allocation/free and tree copy to be very efficient.  E.g., a 64K entries of GNU order-statistics tree copy takes about 6ms, while the code takes about <1ms

- Pointers are replaced by node index, which saves the tree size and may be more friendly to cache

Performance comparison:
ACCESS_ROUND = 64
- Before: 110 H/s
- After: 1200 H/s

ACCESS_ROUND = 128
- Before: 92 H/s
- After: 640 H/s

### DIFF
--- a/qkchash/Makefile
+++ b/qkchash/Makefile
@@ -1,13 +1,16 @@
 CC=g++
 CFLAGS=-Wall -O3 -std=c++17
 
-all: qkchash libqkchash.so
+all: qkchash libqkchash.so qkchash_llrb
 
-qkchash: qkchash.cpp
+qkchash: qkchash.cpp util.h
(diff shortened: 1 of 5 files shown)


qizhou:
Add util.h in the new code.

ninjaahhh:
@qizhou should we update this PR and merge into master?

qizhou:
need to rebase and update the latest code.  Will do it later.

qizhou:
The fix is in #697.  Close it as dup.

PR #215, opened 2018-11-29
Add a template cluster config for testnet 2.0

qcgg:
with guardian public key and bootnodes added


qcgg:
we need to provide instructions or a cmd line tool for users to set coinbase address, private key for p2p, etc...

qcgg:
guardian private key is here https://github.com/QuarkChain/quarkchain-dev/blob/master/deploy/testnet/GUARDIAN.md

PR #217, opened 2018-11-30
optimize cache generation to be more consistent

qizhou:
Simplify cache generation and print some hashes of test vectors

### DIFF
--- a/qkchash/qkchash.py
+++ b/qkchash/qkchash.py
@@ -54,17 +54,15 @@ def fnv64(v1, v2):
 
 
 def make_cache(entries, seed):
-    o = sha3_512(seed)
     cache = []
     cache_set = set()
 
-    for i in range(entries // len(o)):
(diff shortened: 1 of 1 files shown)


qkchash/qkchash.py
```diff
@@ -263,5 +261,18 @@ def test_qkchash_perf():
     print("Equal: ", h0 == h1[0 : len(h0)])
 
 
+def print_test_vector():
+    cache = make_cache(CACHE_ENTRIES, bytes())
+    print("Cache size: ", len(cache))
+    print("Hash of empty:")
+    h0 = qkchash(bytes(), bytes(), cache)
+    print(deserialize_hash(h0["mix digest"]))
```
qcdll: consider adding the result to unittest cases?

qizhou: Nice suggestion!

PR #218, opened 2018-11-30
Sign root block with guardian privkey if specified

ninjaahhh:
1. sign with guardian private key during PoW submission
1. also lower diff if guardian private is specified in config. note that a fake private key can only pass work submission but not blockchain appending, which is part of consensus


quarkchain/config.py
```diff
@@ -233,7 +233,10 @@ class QuarkChainConfig(BaseConfig):
     BLOCK_EXTRA_DATA_SIZE_LIMIT = 1024
 
     PROOF_OF_PROGRESS_BLOCKS = 1
-    GUARDIAN_PUBLIC_KEY = "00" * 64  # TODO: update this
+    GUARDIAN_PUBLIC_KEY = "ab856abd0983a82972021e454fcf66ed5940ed595b0898bcd75cbe2d0a51a00f5358b566df22395a2a8bf6c022c1d51a2c3defe654e91a8d244947783029694d"
+    GUARDIAN_PRIVATE_KEY = (
+        None
+    )  # at the early state, only specified in nodes owned by QuarkChain team
```
qcdll: only nodes certified by QuarkChain team

ninjaahhh:
ah this seems wrong. should sign the block during adding, instead of creating

PR #247, opened 2018-12-04
Remove adaptive guardian diff adjustment

ninjaahhh:
will keep dividing diff by 1000. for mainnet, will investigate more into the actual parameters for adapting guardian difficulty


ninjaahhh:
will **not** merge until next network restart

qcdll:
we should merge this to a new branch (eg develop) as the new release candidate

PR #250, opened 2018-12-05
Update Dockerfile

TahiG:
removed typos


ninjaahhh:
thanks for checking, but it's intended to install `librocksdb-dev` from package manager instead of building from source since the latter takes more time. 

the down side of not building from source is the lack of lz4 compression support but we think it's fine for now (which means in the following snippet try-block will fail and then fall back to snappy)

https://github.com/QuarkChain/pyquarkchain/blob/30da626ef6b50bd07b230aac05a3cef7fd7f23cf/quarkchain/db.py#L73-L79

TahiG:
My bad, hope I didn't offend. Carry on! :)

TahiG:
I've just run it with my 'corrections' and it **is** taking forever.

ninjaahhh:
no problem :)
from our previous experiences building rocksdb from source roughly takes about 20 minutes

PR #251, opened 2018-12-06
add tracking of objects in memory

qcdll:
dead simple, let's see if this reveals something
I tried a few options to track pypy memory usage/leak, eg tracemalloc, vmprof
but in the end only objgraph seems to work OK:
https://stackoverflow.com/questions/45210681/how-do-i-profile-memory-usage-with-pypy

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -1493,6 +1493,12 @@ def parse_args():
 
     return env
 
+async def print_objgraph_growth(env):
+    import objgraph
+    while True:
+        Logger.warning(objgraph.growth())
+        await asyncio.sleep(60)
+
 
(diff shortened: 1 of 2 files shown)


qizhou:
Wohu!!!

ninjaahhh:
oh, realized we probably don't need to merge this.

qcdll:
well, this looks cheap so why not
===
ran this on some cluster for a few hours and master.py memory grew to 1+gb
output is https://gist.github.com/3bf8d3c73ea6e5b62af0c44239fa196a
weakref looks suspicious

qcdll:
this did not uncover the real memory leak (weakref is red herring)

so not going to merge this

Issue #252, opened 2018-12-07
memory leak


qcdll:
master.py has memory leak, looking into it

PR #253, opened 2018-12-07
fix keccak image memory leak

qcdll:
#252 
this causes tests to break now, because of the `copy()` call in tests, will fix later


qcdll:
tests fixed

qcdll:
verified:
1. connects to testnet fine
2. memory leakage from preimage is gone

PR #254, opened 2018-12-07
use tracemalloc and vmprof to track mem usage

qcdll:
FYI purpose, not going to merge

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -1493,6 +1493,33 @@ def parse_args():
 
     return env
 
+async def tracemalloc_snapshot(env):
+    import tracemalloc
+    tracemalloc.start(25)
+    snapshot_old = None
+
+    # vmprof ***
+    import vmprof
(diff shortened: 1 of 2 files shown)


qcdll:
in the end, tracemalloc's compare_to() in snapshot helped identify the root cause of memory leak in #252 

we can use this PR in the future if memory issue arises again

qcdll:
in addition to objgraph, tracking size of objects:
```
slave.add_block_futures
shard.state.new_block_pool
shard.add_block_futures
shard.state.db.m_header_pool
shard.state.db.m_meta_pool
shard.state.db.x_shard_set
shard.state.db.r_header_pool
shard.state.db.r_minor_header_pool
shard.state.db.height_to_minor_block_hashes
```

qcdll:
looks `shard.state.db.height_to_minor_block_hashes` has steady increase over time but the count is not out of ordinary, will revisit with new testnet 23

PR #256, opened 2018-12-07
Check diff match first in ethash validation

ninjaahhh:
before:

`check_pow` pass: ~3ms
`check_pow` fail: ~3ms

after:

`check_pow` pass: ~3.5ms
`check_pow` fail: ~25µs

maybe not worth the change

### DIFF
--- a/ethereum/pow/ethash.py
+++ b/ethereum/pow/ethash.py
@@ -1,6 +1,6 @@
 import copy
 from functools import lru_cache
-from typing import Callable, Dict, List
+from typing import Callable, Dict
 
 from ethereum.pow.ethash_utils import *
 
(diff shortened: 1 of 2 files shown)


qcdll:
this could be applied to shard `handle_new_block`
https://github.com/QuarkChain/pyquarkchain/blob/bb54e7970c8075ce7d9ba515ae3a9109abca0cf1/quarkchain/cluster/shard.py#L526
instead of calling validate_seal we just check difficulty (as in this PR) before forwarding the block to peers.

This may further help block propagation speed by broadcasting sooner once a block is received. But then make the network more susceptible to spam attacks, where a malicious party generate spam blocks by crafting nonce+mixhash to satisfy difficulty (the problem that the attacker is trying to solve would be 2x sha256 which is a lot easier than ethash/qkchash)

I haven't done research on the feasibility of such attack given the difficulty level of mature (eg eth) blockchains, so given this concern, we are not applying the concept here to block propagation

ninjaahhh:
close as the benefit is marginal

PR #274, opened 2018-12-11
Simplify code

qcgg:


quarkchain/cluster/master.py
```diff
@@ -310,12 +310,9 @@ def _task_to_dict(peer, header):
 
     def _pop_best_task(self):
         """ pop and return the task with heightest root """
-        task = None
-        for peer, header in self.tasks.items():
-            if not task or task[0].height < header.height:
-                task = (header, peer)
+        peer, header = max(self.tasks.items(), key=lambda pair: pair[1].height)
```
qcdll: assert tasks is not empty?

ninjaahhh: yes, `max([])` will throw

ninjaahhh: but line 319 makes sure it won't happen. but `assert` won't hurt

PR #276, opened 2018-12-12
Clean up util scripts

ninjaahhh:
remove some unused scripts in `quarkchain/testnet` and move others to `quarkchain/tools`


ninjaahhh:
@qcgg `quarkchain/testnet/block_owners.py → quarkchain/tools/block_owners.py`

will this path change break some thing? I thought you are running this script as a cronjob?

PR #280, opened 2018-12-13
Use host instead of ip for slave config

ninjaahhh:
so we can specify slave host as `qkc-{0..N-1}.cluster.default.svc.cluster.local`, needed during pod discovery where it's harder to get / assign static IP in advance

tested locally, both specifying host and ip work


qcdll:
makes sense, I still suggest merging to a new branch other than master to avoid breaking people wanting to join testnet before the new one is launched

PR #286, opened 2018-12-14
Quicker miner tools--- add a new screen terminal

jyouyj:
I add a new screen terminal for external miners such that the miner can revisit the previous mining logs.


qizhou:
Any updates on this diff?

jyouyj:
Quicker miner tool is suitable for testnet 2.4 and can revisit the previous mining logs.

PR #290, opened 2018-12-15
Shutdown slave server properly when connection closed

ninjaahhh:
also make slave and master server logic look more similar

### DIFF
--- a/quarkchain/cluster/slave.py
+++ b/quarkchain/cluster/slave.py
@@ -897,8 +897,6 @@ class SlaveServer:
         if not self.shutdown_future.done():
             self.shutdown_future.set_result(None)
 
-        # if self.master is not None:
-        #     self.master.close()
         self.slave_connection_manager.close_all()
         self.server.close()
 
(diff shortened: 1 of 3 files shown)


quarkchain/cluster/slave.py
```diff
@@ -887,26 +887,23 @@ def __get_shard_size(self):
...
     def shutdown(self):
-        if self.shutdown_in_progress:
-            return
+        if not self.shutdown_future.done():
+            self.shutdown_future.set_result(None)
 
-        self.shutdown_in_progress = True
-        if self.master is not None:
-            self.master.close()
+        # if self.master is not None:
```
ninjaahhh: in my understanding, shutdown is usually caused by master connection error, which means `MasterConnection.close` will invoke `SlaveServer.shutdown`. amiright? safe to remove these two lines? (otherwise infinite recursion

ninjaahhh: @qcgg

quarkchain/protocol.py
```diff
@@ -266,7 +266,6 @@ def __init__(
         """
         size_bytes = await self.__read_fully(4, allow_eof=True)
         if size_bytes is None:
-            self.close()
```
ninjaahhh: callsite (line 178, same file, `loop_once` func) will also close

PR #291, opened 2018-12-15
Add a sync only tool

HangyuYe:
Just remove miner_address code from quick miner tools script.
This sync tool can ignore fill out address when user launch cluster each times.

When user launch miner first time should use miner tool

Otherwise just use sync tool


qcdll:
looks fine, but what is the use case of this tool?

also, it would be better to change the original tool to add a way to bypass the address input rather than copying portions of code out https://github.com/QuarkChain/pyquarkchain/blob/master/quarkchain/tools/quick_miner_tool.sh

jyouyj:
I will add it.

PR #297, opened 2018-12-21
multi native token id encoding

qcdll:
https://leetcode.com/problems/excel-sheet-column-title/
https://leetcode.com/problems/excel-sheet-column-number/


qcdll:
cross-checked with leetcode online judge, passed all tests

qcdll:
just leaving this as is to match our current application 
can easily generalize in the future for larger id space

PR #300, opened 2018-12-22
undo proof-of-progress

qcgg:


ninjaahhh:
hmm why there is no CI for this PR

qcgg:
I can see it
https://circleci.com/gh/QuarkChain/pyquarkchain/2329?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

ninjaahhh:
what's the best way to release this change? since without updating all clients, some of them will be working in forks

qcdll:
let's create a new branch (development) and merge all breaking changes to it, instead of merging to master

qcgg:
we can do that but at some point we have to launch a tesnet with those new features. they have to be tested before going into mainnet.

qcdll:
as soon as we launch a new testnet (2.4), merge this development branch into master, and keep iterating like that

ninjaahhh:
doesn't seem like it but just to be safe: does empty root block break any of our assumptions?

qizhou:
empty does not break our Bonson consensus, but need to double check the
existing code to make sure pop is disabled.

qcgg:
this change is actually safe to be pushed out as long as the guardian node doesn't have this. :)

quarkchain/cluster/shard_state.py
```diff
@@ -1112,74 +1113,87 @@ def add_root_block(self, root_block: RootBlock):
...
+            # the current header_tip might point to a root block on a fork with r_block
+            # we need to scan back until finding a minor block pointing to the same root chain r_block is on.
+            # the worst case would be that we go all the way back to orig_block (shard_header)
+            while not self.__is_same_root_chain(
+                self.root_tip,
+                self.db.get_root_block_header_by_hash(
+                    self.header_tip.hash_prev_root_block
+                ),
+            ):
+                self.header_tip = self.db.get_minor_block_header_by_hash(
```
qizhou: We may need to call  __rewrite_block_index_to() to header_tip to make sure db is consistent with the tip.

PR #301, opened 2018-12-24
add proof_of_staked_work.py

qizhou:
Add a simulator for proof of staked work.  The basic idea is that a miner needs to lock some percentage of stakes (proportional to its hash power percentage to the network) before producing the blocks.  Rather than choosing PoW or PoS, the consensus needs both PoW and PoS and try to inherit their benefits
- Security bring by PoW
- Potential energy saving by PoS

I will draft an article about this idea asap.

### DIFF
--- a/quarkchain/experimental/proof_of_staked_work.py
+++ b/quarkchain/experimental/proof_of_staked_work.py
@@ -1,0 +1,59 @@
+import random
+
+# Hash power of each miner
+# h = [100, 100, 100, 100]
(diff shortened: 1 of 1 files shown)


qcgg:
Idea is nice - the number of blocks that could be mined by a miner is bounded by the stakes they hold.
Some questions

This would add variance to the block time as some random amount of hash power will be lost in each round if the stake is unpropotional to the hash power?

How would this save energy?

Issue #306, opened 2019-01-08
[RFC] Implement PoSW


ninjaahhh:
a quick eng doc on planned implementation for [Proof of Staked Work](https://medium.com/quarkchain-official/proof-of-staked-work-ef36f9499279) (PoSW) on pyquarkchain.

******

for details on parameter definition please check the article linked above.

consensus parameters:

1. `posw_alpha`: the coefficient for calculating effective hash power
1. `posw_beta`: block difficulty multiplier if already produced enough blocks
1. `posw_window_size`: block range for estimating effective hashpower percentage
1. (temporary and testnet only) `posw_total_stakes`: for simplicity, will assume the amount of total stakes is a constant. a more reasonable version is a constant about the number of stakes needed to produce one block in the window.

UPDATE: combine alpha and total stakes to one param.

extra notes:

1. to better test this new consensus algo, may adopt different parameters in different shards so that we can observe different behaviors to further help parameter tuning
******
consensus control flow change:

i. on validating tx

fail a tx in VM level if the tx.sender is not allowed

ii. on validating block

1. when a new block is going to be added, look up previous `posw_window_size` blocks, count how many blocks are produced by the current block's coinbase address (`produced_blocks`)
1. count amount of tokens in the coinbase address (`stake`), calculate `stake_ratio = min(stake / posw_total_stakes, 1)`,
    1. if `produced_blocks / posw_window_size <= stake_ratio * posw_alpha`, accept the block
    2. otherwise the effective difficulty should be multiplied by `posw_beta`, and if the block nonce doesn't meet the requirement, reject the block; otherwise accept the block

qizhou:
4, may be replaced by a constant about the number of stakes needed to produce one block in the window.

qcgg:
`produced_blocks / posw_window_size <= stake_ratio * posw_alpha`?

PR #313, opened 2019-01-14
[posw p0] Get coinbases block cnts during PoSW window

ninjaahhh:


ninjaahhh:
using block as the argument for `_get_posw_coinbase_balances` seems worse (because I want to invoke it during `add_tx` as well which doesn't have a block)

will change it to use `height` and `evm_state` as arguments (a little bit more tricky to make it work with `lru_cache`)

quarkchain/cluster/shard_state.py
```diff
@@ -85,6 +86,8 @@ def __init__(self, env, full_shard_id: int, db=None, diff_calc=None):
 
         # new blocks that passed POW validation and should be made available to whole network
         self.new_block_pool = dict()
+        # height -> [coinbase address] during previous blocks, ascending order
+        self.coinbase_addr_cache = dict()  # type: Dict[int, Deque[bytes]]
```
qcgg: this cache will break when there is a fork - multiple blocks with the same height

qcgg: #319 has the same problem. we should use block header hash than height.

ninjaahhh: good catch. updated.

quarkchain/cluster/shard_state.py
```diff
@@ -85,6 +86,8 @@ def __init__(self, env, full_shard_id: int, db=None, diff_calc=None):
 
         # new blocks that passed POW validation and should be made available to whole network
         self.new_block_pool = dict()
+        # header hash -> (height, [coinbase address]) during previous blocks (ascending)
+        self.coinbase_addr_cache = dict()  # type: Dict[bytes, Tuple[int, Deque[bytes]]]
```
qizhou: I would suggest using lrucache in https://docs.python.org/3/library/functools.html

quarkchain/config.py
```diff
@@ -110,6 +110,13 @@ class POWConfig(BaseConfig):
     REMOTE_MINE = False
 
 
+class POSWConfig(BaseConfig):
+    STAKE_COEFF = 20  # Alpha in PoSW
```
qizhou: I think STAKE_PER_BLOCK may be easier

quarkchain/cluster/shard_state.py
```diff
@@ -1523,3 +1508,49 @@ def _get_evm_state_from_height(self, height: Optional[int]) -> Optional[EvmState
...
+        if len(self.coinbase_addr_cache) > 128:  # size around 640KB if window size 256
+            self.coinbase_addr_cache = {
+                k: (h, addrs)
+                for k, (h, addrs) in self.coinbase_addr_cache.items()
+                if h > height - 16  # keep most recent ones
+            }
+        return list(addrs)
+
+    @functools.lru_cache(maxsize=16)
+    def _get_posw_coinbase_blockcnt(self, header_hash: bytes) -> Dict[bytes, int]:
```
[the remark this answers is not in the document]
ninjaahhh: a more generalized method `__get_coinbase_addresses_until_block` is provided in case similar functionalities are needed (which doesn't include any PoSW-related param). this method simply serves as the wrapper over that.

quarkchain/cluster/shard_state.py
ninjaahhh: wrap `functools.lrucache` to the method (`__get_coinbase_addresses_until_block`) is tricky in this use case since we have a rolling window AND we want to limit how recursive the function goes:

`rolling_window(n) = rolling_window(n-1).popleft().append(n)`

in our happy path, the simple dict-based cache should work well since blocks will be appended one by one.

fyi the wrapper method `_get_posw_coinbase_blockcnt` indeed uses `lrucache`

quarkchain/config.py
ninjaahhh: discussed offline, will combine alpha and total stakes to a single parameter

quarkchain/cluster/shard_state.py
```diff
@@ -1523,3 +1508,49 @@ def _get_evm_state_from_height(self, height: Optional[int]) -> Optional[EvmState
...
+            _, addrs = self.coinbase_addr_cache[prev_hash]
+            addrs = addrs.copy()
+            if len(addrs) == length:
+                addrs.popleft()
+            addrs.append(header.coinbase_address.recipient)
+        else:  # miss, iterating DB
+            addrs = deque()
+            for _ in range(length):
+                addrs.appendleft(header.coinbase_address.recipient)
+                header = self.db.get_minor_block_header_by_hash(
```
qcgg: add a check?
```
if header.height  == 0:
    break
header = self.db.get_minor_block_header_by_hash(
    header.hash_prev_minor_block
)
check(header is not None)
```

ninjaahhh:
part of #306

PR #319, opened 2019-01-15
[posw p1] Implement lock-up for PoSW coinbases

ninjaahhh:
also using `POSW_CONFIG.ENABLED` as the feature flag


quarkchain/cluster/shard_state.py
```diff
@@ -671,7 +690,8 @@ def add_block(self, block, skip_if_too_old=True):
                     )
                 )
 
-        if self.db.contain_minor_block_by_hash(block.header.get_hash()):
+        header_hash = block.header.get_hash()
```
qizhou: I don't prefer pre-mature optimization, especially such variable is used multiple places in the method and it is hard to find its context when reading its references.

quarkchain/cluster/shard_state.py
```diff
@@ -119,36 +120,48 @@ def init_from_root_block(self, root_block):
...
         self.__rewrite_block_index_to(
-            self.db.get_minor_block_by_hash(self.header_tip.get_hash()),
-            add_tx_back_to_queue=False,
+            self.db.get_minor_block_by_hash(header_tip_hash), add_tx_back_to_queue=False
         )
 
-    def __create_evm_state(self):
-        return EvmState(
+    def __create_evm_state(
+        self, root_hash: Optional[bytes], header_hash: Optional[bytes]
```
qizhou: I would suggest renaming root_hash to trie_root_hash to avoid confusion with root block hash

qizhou: or evm_state_root_hash

quarkchain/evm/messages.py
```diff
@@ -366,6 +366,7 @@ def __init__(self, state, tx):
         self.reset_storage = state.reset_storage
         self.tx_origin = tx.sender if tx else b"\x00" * 20
         self.tx_gasprice = tx.gasprice if tx else 0
+        self.posw_disallow_list = state.posw_disallow_list
```
qcgg: nit:
following the same pattern here, we can add a `is_locked` function to state and then pass that function to VMExt.

ninjaahhh: I think it's more of a functional approach, and mostly a coding style thing. personally I think passing the list to VM is rather more straightforward and easier to read (unless in the future we would have a more complicated logic on disallowing senders)

ninjaahhh:
part of #306

PR #320, opened 2019-01-16
[native] transfer tokens in shard state and evm

qcdll:
all tests passing, design https://github.com/QuarkChain/pyquarkchain/wiki/Multi-Native-Token-in-QuarkChain

there are 2 restrictions on using native tokens in a transaction:
1. **gas_token_id can only be DEFAULT_TOKEN (id=0)**: to support native gas, we need to modify block data structure to store taxes amount in native token
2. **when calling smart contract, transfer_token_id has to be DEFAULT_TOKEN (id=0)**: to support other native tokens, evm needs some change (also need to change BALANCE opcode)

### DIFF
--- a/quarkchain/cluster/rpc.py
+++ b/quarkchain/cluster/rpc.py
@@ -466,8 +466,8 @@ class GetAccountDataRequest(Serializable):
         self.block_height = block_height
 
 
-def token_pair_list_to_dict(list):
-    return {p.token_id: p.balance for p in list}
+def token_pair_list_to_dict(l):
+    return {p.token_id: p.balance for p in l}
 
 
 class TokenBalancePair(Serializable):
--- a/quarkchain/cluster/rpc.py
+++ b/quarkchain/cluster/rpc.py
@@ -466,18 +466,30 @@ class GetAccountDataRequest(Serializable):
         self.block_height = block_height
 
 
+def token_pair_list_to_dict(list):
+    return {p.token_id: p.balance for p in list}
+
+
+class TokenBalancePair(Serializable):
+    FIELDS = [("token_id", uint64), ("balance", uint256)]
+
+    def __init__(self, token_id, balance):
+        self.token_id = token_id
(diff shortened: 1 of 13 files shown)


quarkchain/evm/state.py
```diff
@@ -317,8 +322,15 @@ def get_and_cache_account(self, address):
...
-        return self.get_and_cache_account(utils.normalize_address(address)).balance
+    def get_balances(self, address):
+        return self.get_and_cache_account(
+            utils.normalize_address(address)
+        ).token_balances.balances
+
+    def get_token_balance(self, address, token_id):
+        return self.get_and_cache_account(
+            utils.normalize_address(address)
+        ).token_balances.balance(token_id)
```
ninjaahhh: should be `balance[token_id]`?

qcdll: `balance()` is a function, either that or `balances[token_id]`, but the latter does not handle the case where the token in question has been added to the account before

quarkchain/cluster/rpc.py
```diff
@@ -466,18 +466,30 @@ def __init__(self, address: Address, block_height: typing.Optional[int] = None):
         self.block_height = block_height
 
 
+def token_pair_list_to_dict(list):
```
ninjaahhh: avoid reserved keyword `list`

quarkchain/evm/state.py
```diff
@@ -114,17 +116,19 @@ def serialize(self):
         return retv
 
     def balance(self, token_id):
-        self.balances.get(token_id, 0)
+        return self.balances.get(token_id, 0)
 
-    def delta(self, token_id, value):
-        self.balances[token_id] = self.balances.get(token_id, 0) + value
+    def is_empty(self):
+        for k, v in self.balances.items():
```
ninjaahhh: `return all(v == 0 for v in self.balances.values())`

PR #327, opened 2019-01-18
[native] fix SUICIDE opcode

qcdll:
correctly transfer all destructed contract token balances to the designated address

this was a missed point when changing to support native tokens. Seems no test is covering this OPCODE, will see if we can add some


qcdll:
discussed offline, will only send DEFAULT TOKEN to designated address when calling suicide
ALL OTHER NATIVE TOKENS AT CONTRACT ADDRESS WILL NOT BE ACCESSIBLE

PR #336, opened 2019-01-23
[native] get rid of DEFAULT_TOKEN

qcdll:
replace with genesis_token or default_chain_token

also, use default_chain_token in EVM, so that there is no change to EVM or smart contract and they will be able to interact with multi native token, as long as the default_chain_token is set

see design rationale in https://github.com/QuarkChain/pyquarkchain/wiki/Multi-Native-Token-in-QuarkChain


quarkchain/evm/state.py
```diff
@@ -369,7 +370,16 @@ def set_nonce(self, address, value):
...
     def set_token_balance(self, address, token_id, val):
         acct = self.get_and_cache_account(utils.normalize_address(address))
-        if val == self.get_token_balance(address, token_id):
+        if val == self.get_balance(address, token_id=token_id):
+            self.set_and_journal(acct, "touched", True)
+            return
+        self._set_token_balance_and_journal(acct, token_id, val)
+        self.set_and_journal(acct, "touched", True)
+
+    def set_balance(self, address, val):
```
qcgg: call set_token_balance?

quarkchain/evm/state.py
```diff
@@ -263,6 +261,7 @@ def __init__(
         self.executing_on_head = executing_on_head
         self.qkc_config = qkc_config
         self.sender_disallow_list = set()  # type: Set[bytes]
+        self.shard_config = ShardConfig(ChainConfig())
```
ninjaahhh: I feel there is too much abstraction leaking from shard to evm state, ideally the evm state should not know what is a shard / chain. how about just pass `default_chain_token` to the constructor?

qcdll: what if we want to add a permitted token list in EVM? how about letting the EVM know which shard it belongs to and what contract from neighbor it can talk to? those are examples of what we can do with a more complete picture of shard config, rather than a default_chain_token

having more info than needed is better considering we might get more features into EVM soon

ninjaahhh: it's no big deal for now. when the config gets more complicated we can refactor to something like `EVMConfig`.

one final note, can you pass the config to the constructor and make it required? otherwise it's possible that the caller forgot to update the `self.shard_config` in place. or make `self.shard_config` to `None` so unexpected invocations will fail immediately

qcdll: we already have quarkchain.evm.config.Env for this purpose

if the caller forgot to update the shard_config, it would use the default value which is the current code, the EVM would have a default value to use

what benefit does failing the invocations have? besides making our untittests fail

qcdll: the hard part is picking an initial/default value for the default token for EVM, using either 0 or "TQKC" is bad. the cleanest way of reusing whatever we defined in ShardState is going to be something like this. I can add an optional argument to `State.__init__` but it does not really help people remember to adjust ShardState, given the argument list is quite long and even contains kwargs

qcdll:
discussed offline we'll keep this as is now

PR #337, opened 2019-01-24
store minor coinbase in root state

### LINKED ISSUE #324: need to verify minor block coinbase when processing root blocks
https://github.com/QuarkChain/pyquarkchain/blob/0d0bb3144625970490d10d42a362c58e6c3f4d4f/quarkchain/cluster/root_state.py#L400
this TODO needs addressing

qcdll:
one step closer to removing coinbase_amount in RootBlockHeader and MinorBlockHeader, and allowing multi native token as coinbase rewards (both root block and minor block)

when minor blocks are validated and added to shard states, a mapping (also stored in DB) will be available in root_state with corresponding coinbase amounts

this fixes numerous issues in test_root_state.py as coinbase calculations were incorrect; also, fixes #324

### DIFF
--- a/quarkchain/cluster/root_state.py
+++ b/quarkchain/cluster/root_state.py
@@ -15,8 +15,6 @@ from quarkchain.core import (
     Serializable,
     calculate_merkle_root,
     TokenBalanceMap,
-    TokenBalancePair,
-    TokenBalanceList,
 )
 from quarkchain.diff import EthDifficultyCalculator
 from quarkchain.genesis import GenesisManager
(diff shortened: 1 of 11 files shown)


quarkchain/core.py
```diff
@@ -1026,14 +1024,14 @@ def __init__(
 
     def finalize(
         self,
-        coinbase_amount_map=TokenBalanceMap(dict()),
-        coinbase_address=Address.create_empty_account()
+        coinbase_amount_map=dict(),
```
ninjaahhh: there could be some hidden bugs if the underlying dictionary is modified later. recommended way is to use `None` as default arg and init to a new dict instance if it's `None`

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

qcdll: this (or even `TokenBalanceMap(dict())`) is bad because default param can be modified and affecting all subsequent calls

quarkchain/cluster/root_state.py
```diff
@@ -185,11 +188,15 @@ def get_root_block_by_height(self, height):
 
     # ------------------------- Minor block db operations --------------------------------
     def contain_minor_block_by_hash(self, h):
-        return h in self.m_hash_set
+        return h in self.m_hash_dict.keys()
 
-    def put_minor_block_hash(self, m_hash):
-        self.db.put(b"mheader_" + m_hash, b"")
-        self.m_hash_set.add(m_hash)
+    def put_minor_block_hash(self, m_hash: bytes, coinbase_tokens: dict):
```
ninjaahhh: nit: `put_minor_block_coinbase`? somehow let the reader know key and value

PR #343, opened 2019-02-04
initial version to support collecting different token taxes in rootblock

qizhou:
AddMinorBlockHeader now return TokenBalanceMap, however SyncMinorBlockListRequest needs to be updated to return TokenBalanceMap.


qizhou:
going to close the diff as it is replaced by #344 and #337

PR #344, opened 2019-02-07
rename RootBlockHeader coinbase_amount to coinbase_amount_map

qizhou:
Rename RootBlockHeader coinbase_amount to coinbase_amount_map so that it could collect multiple token id's as tax.  Currently, only accept QKC (tQKC)

### DIFF
--- a/quarkchain/cluster/root_state.py
+++ b/quarkchain/cluster/root_state.py
@@ -376,6 +376,7 @@ class RootState:
             actual_coinbase_amount = block.header.coinbase_amount_map.balance_map.get(
                 self.env.quark_chain_config.genesis_token, 0)
 
+            # TODO:  Support collecting tax from multiple native tokens
             if len(block.header.coinbase_amount_map.balance_map) > 1:
                 raise ValueError("Incorrect coinbase_amount_map: too many tokens")
 
--- a/quarkchain/cluster/root_state.py
+++ b/quarkchain/cluster/root_state.py
@@ -14,6 +14,7 @@ from quarkchain.core import (
     RootBlockHeader,
     Serializable,
     calculate_merkle_root,
+    TokenBalanceMap,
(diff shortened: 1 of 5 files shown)


quarkchain/cluster/root_state.py
```diff
@@ -367,13 +372,25 @@ def validate_block(self, block, block_hash=None):
...
-            coinbase_amount = self.__calculate_root_block_coinbase(block)
-            if coinbase_amount != block.header.coinbase_amount:
+            expected_coinbase_amount = self.__calculate_root_block_coinbase(block)
+            actual_coinbase_amount = block.header.coinbase_amount_map.balance_map.get(
+                self.env.quark_chain_config.genesis_token, 0)
+
+            if len(block.header.coinbase_amount_map.balance_map) > 1:
+                raise ValueError("Incorrect coinbase_amount_map: too many tokens")
+
+            if (
```
ninjaahhh: in the future we also expect multiple tokens in coinbase balance map, right? if so, can add a TODO comment here for this check to clarify

qcgg:
this potentially could grow quite large?

ninjaahhh:
> this potentially could grow quite large?

a quick calc if we have a large amount of different tokens: one minor block can have 8mil/21000 ~= 400 tx with different gas tokens, meaning balance map will have 400 entries max; in current setting a root block can have 32 minor blocks per shard, with 8 shards in total we would expect 400 * 32 * 8 = 100k different tokens in balance map, which resides in the header :(

qizhou:
The should be eventually replaced by the hash of the map.  However,  I
would expect in a first few years, the number of native tokens created will
be less than 10.

If the number of native tokens grows to a large number, we could easily
upgrade it using the hash version, but the current version should be enough
for mainnet launch.

- Qi

PR #348, opened 2019-02-09
use fixed gas limit instead of dynamic one

qizhou:
Use fixed gas limit so that the source shard could impose the proper x-shard gas limit for each x-shard tx so that a x-shard tx won't exceed the target gas limit.


qizhou:
> in the future will the gas limit be different in each shard?

Depending on the consensus and its parameters (such as block interval), the gas limit (or other limit, CPU, BW) may be different.

PR #350, opened 2019-02-10
Improve cross-shard using cursor

qizhou:
This diff implements x-shard transactions using a cursor, where
- Assume xshard and inshard gas limit are fixed for all shards (6M x-shard and 6M in-shard)
- Sender: throttling at sender is removed.
- Sender: any cross-shard tx with startgas exceeding xshard gas limit will be rejected
- Receiver: it will iterate all x-shard deposits until xshard gas limit is reached (soft limit).
- Receiver: cursor data will be stored in block meta so that it can be quickly recovered.
- Misc: x-shard and in-shard limits can be customized for testing


qizhou:
According to the definition in the code, cursor will be either eof or a
valid cursor so there is no boundary check here.

- Qi

PR #356, opened 2019-02-11
add ENABLE_TX_TIMESTAMP config

qizhou:
For bootstrap purpose so that tx will be enabled after a bootstrap period.


ninjaahhh:
nit: similar as in #357, using block height seems more natural

qcgg:
btw, ethereum did this by limiting the block gas.

qizhou:
Similar to EVM case, even we use root block as clock, the minor blocks may not include the root block at the same time, and thus the actual time of enabling TX of all shards are not the same...

PR #357, opened 2019-02-11
add ENABLE_EVM_TIMESTAMP config

qizhou:
Add ENABLE_EVM_TIMESTAMP config so that evm will be disabled before a specific timestamp.


qcgg:
shouldn't we use root block height as the clock?

ninjaahhh:
> shouldn't we use root block height as the clock?

+1, using block height seems more natural

qizhou:
> > shouldn't we use root block height as the clock?
> 
> +1, using block height seems more natural

I tried to use root block height or root block create time, however, __validate_tx() is called without root block information, and passing root block information needs to change a significant amount of code.

Issue #353, opened 2019-02-10
initial data set of QKCHASH should be changed regularly


qizhou:
Similar to ETHASH, QKCHASH's initial data set should be changed for every epoch.

PR #358, opened 2019-02-12
Change Qkchash's initial data set and cache seed for every epoch

jyouyj:
#353 
Similar to ETHASH, QKCHASH's initial data set and cache seed have been changed for every epoch.

Note that this modification changes the consensus, which is incompatible with Testnet 2.4.


qizhou:
What is the progress of this diff?

jyouyj:
Noted that circleCi test does not support qkchash native.

PR #359, opened 2019-02-12
send coinbase token map to master

qcdll:
still want to remove the coinbase field from minor block header, we are getting close to the goal

0. change the minor block header coinbase from int to TokenBalanceMap
1. at the end of `add_block()`, calling `send_minor_block_header_to_master` will also send the coinbase token map
2. at the end of `add_block_list_for_sync()` return a list of coinbase token map and piggyback that into `SyncMinorBlockListResponse`

This eliminates the need for coinbase field in minor block header.

After this, the only place left that still makes use of coinbase field in minor block header is from `get_unconfirmed_headers_coinbase_amount`; there are some tests that uses coinbase_amount_map from minor block header as well, but it can be changed easily; also from some jsonrpc calls

This replaces #343

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -676,7 +676,7 @@ class XshardTxCursorInfo(Serializable):
 
     FIELDS = [
         ("root_block_height", uint64),
-        ("minor_block_index", uint64),              # 0 root block x shard, >= 1 minor blocks
+        ("minor_block_index", uint64),  # 0 root block x shard, >= 1 minor blocks
         ("xshard_deposit_index", uint64),
     ]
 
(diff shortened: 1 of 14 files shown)


quarkchain/core.py
```diff
@@ -801,13 +801,13 @@ def finalize_merkle_root(self):
         self.meta.hash_merkle_root = self.calculate_merkle_root()
         return self
 
-    def finalize(self, evm_state, coinbase_amount, hash_prev_root_block=None):
+    def finalize(self, evm_state, coinbase_amount_map, hash_prev_root_block=None):
```
ninjaahhh: `coinbase_amount_map`  here is of type `TokenBalanceMap`, while in `create_block_to_append` the arg with the same name is a optional dict? can we unify the argument style here?

qcdll: yes very confusing indeed, the latest version uses different names to hint types:
coinbase_amount_map: TokenBalanceMap
coinbase_tokens: Dict

PR #366, opened 2019-02-21
tests for x-shard transfer with non-genesis gas token

qcdll:
make sure that cross shard token transfer with non-genesis token as gas works correctly

### DIFF
--- a/quarkchain/cluster/tests/test_native_token.py
+++ b/quarkchain/cluster/tests/test_native_token.py
@@ -358,8 +358,9 @@ class TestNativeTokenShardState(unittest.TestCase):
         )
 
         # X-shard gas used
-        evmState0 = state0.evm_state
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/tests/test_native_token.py
```diff
@@ -361,6 +361,180 @@ def test_xshard_native_token_received(self):
...
+            state0.get_token_balance(acc3.recipient, self.genesis_token),
+            self.getAfterTaxReward(self.shard_coinbase),
+        )
+        self.assertEqual(
+            state0.get_token_balance(acc3.recipient, QETH),
+            self.getAfterTaxReward(opcodes.GTXXSHARDCOST * 2),
+        )
+
+        # X-shard gas used
+        evmState0 = state0.evm_state
```
ninjaahhh: nit: naming

PR #377, opened 2019-03-01
fixed token supply

qcdll:
block reward decrease by a factor of p (BLOCK_REWARD_DECAY_FACTOR) for every epoch
epoch is defined for each chain, as `block_height // number_of_blocks_per_epoch`

### DIFF
--- a/quarkchain/cluster/root_state.py
+++ b/quarkchain/cluster/root_state.py
@@ -260,14 +260,25 @@ class RootState:
             create_time = max(self.tip.create_time + 1, int(time.time()))
         return self.diff_calc.calculate_diff_with_parent(self.tip, create_time)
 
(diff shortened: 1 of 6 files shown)


ninjaahhh:
assumption is every `quark_chain_config.ROOT.EPOCH_INTERVAL`, number of shards will change **at the same time**?

also the same logic decaying logic will be applied to shard coinbase, as well?

qcdll:
we do not consider number of shards here to minimize cross-dependency and reduce the risk of inconsistency, decaying the coinbase is purely based on epoch (which is based only on block height)

the same logic is already in shard coinbase

qcdll:
https://medium.com/quarkchain-official/fixed-supply-token-model-with-dynamically-added-shards-chains-52db23a35aa3

PR #379, opened 2019-03-05
Add total diff field to root block

ninjaahhh:
so we can update root chain's fork choice rule to be based on total difficulty

cc @pingke

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -570,8 +570,8 @@ class TokenBalanceMap(Serializable):
 
 
 def calculate_merkle_root(item_list):
-    ''' Using ETH2.0 SSZ style hash tree
-    '''
+    """ Using ETH2.0 SSZ style hash tree
+    """
     if len(item_list) == 0:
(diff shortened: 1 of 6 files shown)


quarkchain/core.py
```diff
@@ -856,6 +858,7 @@ class RootBlockHeader(Serializable):
         ("extra_data", PrependedSizeBytesSerializer(2)),
         ("mixhash", hash256),
         ("signature", FixedSizeBytesSerializer(65)),
+        ("total_difficulty", biguint),
```
qizhou: I would suggest moving total_difficulty field just after difficulty field - in general, we want to leave signature to be the last field in the header.

qizhou: I would suggest moving total_difficulty field just after difficulty field - in general, we want to leave signature to be the last field in the header.

ninjaahhh: done

ninjaahhh: done

PR #380, opened 2019-03-06
Use total diff to determine longest root chain

ninjaahhh:
based on #379.


ninjaahhh:
> I guess shard_state.py should also be changed (ShardState.add_root_block()) method.

good call. will update

PR #384, opened 2019-03-11
Remove eth hard forks

ninjaahhh:
we may want to consider removing hard fork checks in ethereum. what do you think? @qizhou cc @pingke


ninjaahhh:
@qizhou can you confirm whether this is needed for singularity launch?

qizhou:
I think we could check it in - as long as we don't support actual tx at the beginning, we have time to check the code and make sure it is consistent with the post-constantinople fork of ETH.

PR #385, opened 2019-03-11
close peer writer properly

### LINKED ISSUE #382: Cluster shutdown due to memory leak
The cluster is shutdown periodically. I think it still has the memory leak problem. I start the cluster yesterday morning, which use about 1.15G memory. When I double check today, it increases to about 2G memory. Since the machine has 15G memory, the cluster can work about two weeks，which matches my observation. Noted that I use the python version instead of pypy3.

qcdll:
see #382, I think the issue is caused by not closing the socket writer properly when a peer is removed from pool
Testplan:
1. follow instructions in `tool_multi_cluster_fd_test.py`
2. before this change, dead peers leave open fds (`lsof -p PID`):
```
pypy3   88790  dll   33u    IPv4 0x60bc88494b75ba93      0t0      TCP localhost:38291->localhost:50377 (CLOSE_WAIT)
pypy3   88790  dll   34u    IPv4 0x60bc88494e630493      0t0      TCP localhost:38291->localhost:50421 (CLOSE_WAIT)
pypy3   88790  dll   35u    IPv4 0x60bc88494135fa93      0t0      TCP localhost:38291->localhost:50483 (CLOSE_WAIT)
pypy3   88790  dll   36u    IPv4 0x60bc88494be1a113      0t0      TCP localhost:38291->localhost:50561 (CLOSE_WAIT)
```
3. after this change, they are gone

will close #382

### DIFF
--- a/quarkchain/cluster/tool_multi_cluster_fd_test.py
+++ b/quarkchain/cluster/tool_multi_cluster_fd_test.py
@@ -1,0 +1,95 @@
+"""
+Direction of use:
+run `python tool_multi_cluster.py --p2p --num_clusters=1`,
+then `python tool_multi_cluster_fd_test.py --p2p --num_clusters=NUM`
+this will launch NUM-1 clusters SEQUENTIALLY, and shut them down after 40 secs, to simulate peers connecting and gone
+"""
+
+
+import argparse
+import asyncio
+import os
+import random
+
+from quarkchain.p2p.utils import colors, COLOR_END
+from quarkchain.cluster import cluster as cl
+from quarkchain.cluster.cluster_config import ClusterConfig
+
+
+async def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


ninjaahhh:
LGTM

quarkchain/cluster/tool_multi_cluster_fd_test.py
```diff
@@ -0,0 +1,95 @@
...
+            cl.Cluster(config, "{}C{}{}_".format(colors[i % len(colors)], i, COLOR_END))
+        )
+
+        args.p2p_port += 1
+        args.port_start += 100
+        args.json_rpc_port += 1
+        args.json_rpc_private_port += 1
+
+    tasks = list()
+    # tasks.append(asyncio.ensure_future(clusters[0].run()))
```
ninjaahhh: nit: remove comment

PR #397, opened 2019-03-13
close connection if handshake fails

qcdll:
this is a bit hacky but verified to fix the `too many open files` issue

### DIFF
--- a/quarkchain/p2p/p2p_server.py
+++ b/quarkchain/p2p/p2p_server.py
@@ -158,10 +158,17 @@ class BaseServer(BaseService):
             await self._receive_handshake(reader, writer)
         except expected_exceptions as e:
             self.logger.debug("Could not complete handshake: %s", e)
+            self.logger.error("Could not complete handshake: %s", e)
+            reader.feed_eof()
+            writer.close()
         except OperationCancelled:
-            pass
+            self.logger.error("OperationCancelled")
+            reader.feed_eof()
+            writer.close()
         except Exception as e:
             self.logger.exception("Unexpected error handling handshake")
+            reader.feed_eof()
+            writer.close()
 
     async def _receive_handshake(
         self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter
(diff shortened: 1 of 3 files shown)


quarkchain/p2p/p2p_server.py
```diff
@@ -158,10 +158,17 @@ def _get_discv5_topic(self) -> bytes:
             await self._receive_handshake(reader, writer)
         except expected_exceptions as e:
             self.logger.debug("Could not complete handshake: %s", e)
+            self.logger.error("Could not complete handshake: %s", e)
```
ninjaahhh: duplicate line

quarkchain/p2p/p2p_server.py
```diff
@@ -158,10 +158,17 @@ def _get_discv5_topic(self) -> bytes:
...
+            reader.feed_eof()
+            writer.close()
         except OperationCancelled:
-            pass
+            self.logger.error("OperationCancelled")
+            reader.feed_eof()
+            writer.close()
         except Exception as e:
             self.logger.exception("Unexpected error handling handshake")
+            reader.feed_eof()
```
ninjaahhh: can use `finally` block

quarkchain/p2p/p2p_server.py
qcdll: they are not duplicate, but I will remove the error log

quarkchain/p2p/p2p_server.py
qcdll: finally will close writer regardless of exception??

ninjaahhh: yup

qcdll: that's not right, shall not close if no exception

ninjaahhh: ah i see. another way to save those duplicate lines would be return at line 158 and close after the try-catch block

qcdll:
actually this was recording opened sockets during handshake, and if the handshake turns out to be unfruitful we close the sockets

at the point of performing handshake, they are not "peers" yet

PR #403, opened 2019-03-13
prevent tx containing token_id not in allowed list from being added

### LINKED ISSUE #413: Should probably prevent arbitrary token
accidentally made a transaction with token TQKC which is not in our testnet config (which is QKC), and it worked

thus ppl can create transactions with arbitrary token to bloat an account's storage? is this a issue we may want to prevent?

qcdll:
test plan:
use old web commit to submit a transaction using `TQKC` as transfer_token_id and gas_token_id, with this PR it fails with:
```
Failed to add transaction: <Transaction(5119)>: 'token 1435440 balance' actual:0 target:'>0'
```

will close #413

### DIFF
--- a/quarkchain/evm/messages.py
+++ b/quarkchain/evm/messages.py
@@ -156,24 +156,11 @@ def validate_transaction(state, tx):
     # (4.0) require transfer_token_id and gas_token_id to be in balances
     # either this, or reject 0 gas_price transactions
     balances = state.get_balances(tx.sender)
-    if tx.transfer_token_id not in balances:
-        raise InsufficientBalance(
-            rp(
-                tx,
-                "token %d balance" % tx.transfer_token_id,
-                state.get_balance(tx.sender, token_id=tx.transfer_token_id),
-                ">0",
+    for token_id in [tx.transfer_token_id, tx.gas_token_id]:
+        if token_id not in balances:
+            raise InsufficientBalance(
+                rp(tx, "token %d balance" % token_id, "non-existent", ">0")
             )
-        )
-    if tx.gas_token_id not in balances:
-        raise InsufficientBalance(
-            rp(
-                tx,
-                "token %d balance" % tx.gas_token_id,
-                state.get_balance(tx.sender, token_id=tx.gas_token_id),
-                ">0",
-            )
-        )
 
     # (4) the sender account balance contains at least the
     # cost, v0, required in up-front payment.
--- a/quarkchain/evm/messages.py
+++ b/quarkchain/evm/messages.py
@@ -25,6 +25,7 @@ from quarkchain.evm.exceptions import (
     InvalidTransaction,
 )
 from quarkchain.evm.slogging import get_logger
+from quarkchain.utils import token_id_decode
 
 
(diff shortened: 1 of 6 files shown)


qcdll:
when we didn't have multi-native token (actually, before we store block reward as a list of tokens), the tx with 0 value and gas will be processed but has no effect on the ledger; I think the tx is still processed and indexed and can be looked up properly

now that we have multiple tokens, it is possible to create a transaction with non-existent token ids, and because no one has that token, gas price or value has to be 0 for the tx to be valid. Then the non-existent token id would appear as part of block reward (if it is used as gas), which is part of block header at the moment. NOTE that we do not store any token id with 0 balance in the state, so storage won't be bloated if someone created lots of tx with non-existent token ids.

With this PR, tx with 0 value or 0 gas will ONLY be rejected if the sender account does not possess any amount of token (need to be >0 to be stored in the state). So for example if you have 100 TQKC, your 0 value TQKC transfer or 0 gas using TQKC tx is still going work. This PR effectively blocks the possibility of non-existent token id appearing in any part of the ledger or blockchain

qizhou:
I see.  Make sense.  Good job here!

quarkchain/evm/messages.py
```diff
@@ -153,6 +153,28 @@ def validate_transaction(state, tx):
     if tx.startgas < total_gas:
         raise InsufficientStartGas(rp(tx, "startgas", tx.startgas, total_gas))
 
+    # (4.0) require transfer_token_id and gas_token_id to be in balances
+    # either this, or reject 0 gas_price transactions
+    balances = state.get_balances(tx.sender)
+    if tx.transfer_token_id not in balances:
```
ninjaahhh: nit

```python
for token_id in (tx.transfer_token_id, tx.gas_token_id):
  if token_id not in balances:
    raise InsufficientBalance(...)
```

qcdll: 👍

qcdll:
unsurprisingly, this caused some other issue: `call` jsonrpc with all-zero address will always fail because of the new restriction in this PR

In the end, we just want to prevent tx containing unknown token id to be persisted into blockchain. The easiest way without hacking around is just to have a whitelisted token ids and invalidate tx that has token ids not in them

qizhou:
> unsurprisingly, this caused some other issue: `call` jsonrpc with all-zero address will always fail because of the new restriction in this PR
> 
> In the end, we just want to prevent tx containing unknown token id to be persisted into blockchain. The easiest way without hacking around is just to have a whitelisted token ids and invalidate tx that has token ids not in them

You mean sending token from or to all-zero address?

qcdll:
no, it's not about sending token from or to all-zero address. `call` jsonrpc will construct all-zero address when no `from` address is given
https://github.com/QuarkChain/pyquarkchain/blob/3f26e929f794cd722f2c736f1cab4aab7019e42c/quarkchain/cluster/jsonrpc.py#L1235
and use this address to make non-persistency calls or estimate gas, if applicable. This PR in original form will invalidate tx from this endpoint. So we moved to a different strategy

Issue #404, opened 2019-03-14
Create a benchmark for QKCHASH multithread


AGx10k:
QKCHASH needs a multithread benchmark tool. I have done it myself but i'm not a coder and the code is very bad. So it's just an issue, not a PR

My code: https://github.com/AGx10k/pyquarkchain/blob/master/qkchash/benchmark-multishread.sh
Test results: https://docs.google.com/spreadsheets/d/1YqbBWfXip1vprj3a53nei1cfogMaz8PW1iDk-Vr9GAo/edit#gid=0

qizhou:
That is a great job here, and many thanks for sharing the result.   From the code, it looks like it is benchmarking python version of QKCHASH, which is much slower than C/C++ version.  Will be great if you could get the results for that (and we could create a PR to add the result in qkchash.MD).

AGx10k:
It's benchmarking the native version [see line 302](https://github.com/AGx10k/pyquarkchain/blob/master/qkchash/benchmark-multishread.sh#L302) in this way:
 - checks that external_miner is not running to be sure the CPU is not loaded
 - cpu_cores_total=count of CPU threads available
 - creates py file which is a slightly modified copy of [qkchash.py](https://github.com/AGx10k/pyquarkchain/blob/master/qkchash/qkchash.py)
 - simultaneously runs $cpu_cores_total copies of py file each logging it's output to separate log file
 - when all copies are finished (checks pgrep) it gets the last lines of output to get the average result

did results myself with some help from of https://t.me/QuarkChain_Russia

here is PR: #410

AGx10k:
but honestly my benchmark is very inefficient and i hope somebody will make a better one which will be more useful.

PR #410, opened 2019-03-15
Create qkchash.md with benchmark results

AGx10k:
benchmark results mentioned in #404


ninjaahhh:
this looks very interesting cc @qizhou

qizhou:
I would suggest adding more details including which code (python or C/C++) and the parameters of QKCHASH.

ninjaahhh:
close as we already updated qkchashx. also suggest putting this to a wiki or an issue then to the wiki

PR #414, opened 2019-03-15
Load stale root block from DB

### LINKED ISSUE #412: Node shuts down when adding stale root block after reboot
reproduction in testnet2.5:

1. start syncing from scratch, stop around root block 520 by terminating `cluster.py` process
2. restart

expect to see errors like following:

```
SLAVE_S0: E0313 03:00:12.366294 protocol.py:168] Traceback (most recent call last):
SLAVE_S0: File "/code/pyquarkchain/quarkchain/protocol.py", line 166, in __internal_handle_metadata_and_raw_data
SLAVE_S0: await self.handle_metadata_and_raw_data(metadata, raw_data)
SLAVE_S0: File "/code/pyquarkchain/quarkchain/cluster/protocol.py", line 83, in handle_metadata_and_raw_data
SLAVE_S0: await super().handle_metadata_and_raw_data(metadata, raw_data)
SLAVE_S0: File "/code/pyquarkchain/quarkchain/protocol.py", line 152, in handle_metadata_and_raw_data
SLAVE_S0: await self.__handle_rpc_request(op, cmd, rpc_id, metadata)
SLAVE_S0: File "/code/pyquarkchain/quarkchain/protocol.py", line 127, in __handle_rpc_request
SLAVE_S0: resp = await handler(self, request)
SLAVE_S0: File "slave.py", line 209, in handle_add_root_block_request
SLAVE_S0: switched = await shard.add_root_block(req.root_block)
SLAVE_S0: File "/code/pyquarkchain/quarkchain/cluster/shard.py", line 505, in add_root_block
SLAVE_S0: return self.state.add_root_block(root_block)
SLAVE_S0: File "/code/pyquarkchain/quarkchain/cluster/shard_state.py", line 1320, in add_root_block
SLAVE_S0: shard_header.hash_prev_root_block
SLAVE_S0: File "/code/pyquarkchain/quarkchain/cluster/shard_state.py", line 544, in __is_same_root_chain
SLAVE_S0: if shorter_block_header.height > longer_block_header.height:
SLAVE_S0: AttributeError: 'NoneType' object has no attribute 'height'
```

deleting all data files and restarting will fix

******

for given lines of [code](https://github.com/QuarkChain/pyquarkchain/blob/a430cfc176784834238733184bd7108152a263f6/quarkchain/cluster/shard_state.py#L1298-L1323)

```python
last_minor_header_in_prev_root_block = self.db.get_last_confirmed_minor_block_header_at_root_block(
    root_block.header.hash_prev_block
)
if shard_headers:
    # Master should assure this check will not fail
    check(
        shard_headers[0].height == 0
        or shard_headers[0].hash_prev_minor_block
        == last_minor_header_in_prev_root_block.get_hash()
    )
    shard_header = shard_headers[-1]
else:
    shard_header = last_minor_header_in_prev_root_block

# shard_header can be None meaning the genesis shard block has not been confirmed by any root block

self.db.put_root_block(root_block, shard_header)
if shard_header:
    check(
        self.__is_same_root_chain(
            root_block.header,
            self.db.get_root_block_header_by_hash(
                shard_header.hash_prev_root_block
            ),
        )
    )
```

this happens when `shard_header == last_minor_header_in_prev_root_block` and `self.db.get_root_block_header_by_hash(shard_header.hash_prev_root_block)` returns `None`.

in another word, the slave doesn't have a root block that's pointed by a minor block which is confirmed by a given root blocks's previous block, in ~database~ memory.

******

from @qcgg 

Here is what might have happened:
`last_minor_header_in_prev_root_block` was really old, so old that its prev_root_block was more than `quark_chain_config.ROOT.max_root_blocks_in_memory` blocks before the current root chain head.
Since we only recover at most `quark_chain_config.ROOT.max_root_blocks_in_memory` root block headers in ShardState after restart, when shard chain hasn't produced a block for a long time, a restart would trigger this check failure.

In this case we should remove consistency check when looking up root block header. `get_root_block_header_by_hash(h, consistency_check=False)`

ninjaahhh:
should fix #412 

context: 

for both root db and shard db, `get_root_block_header_by_hash` by default only reads from memory, unless `consistency_check` flag is false, and the reason is we can only assume those root blocks in memory have consistent view of the network (no missing previous block, missing minor block, etc), while root blocks in DB could be forks with missing information.

however, for nodes who experienced a reboot and state recovery, it's possible for a shard's tip (or some other minor block which we are sure to have a consistent view) to point to a very old root block which is not recovered and loaded into memory, thus `get_root_block_header_by_hash` will fail. 

this PR updates several aforementioned cases, where we would directly load root blocks from database:

1. when adding minor block
  a. loading root block pointed from the shard's tip
2. when adding root blocks
  a. loading root block pointed from the shard's tip
  b. loading root block pointed from the last minor block confirmed by the given root block's predecessor
3. when comparing whether root blocks are of the same chain
  a. loading root blocks between the shorter and the longer
4. get next position for xshrad deposit cursor (@qizhou may need your eye on this one)
  a. load root blocks pointed by the given root block's included minor blocks

the testing scenario is a little bit more tricky since we need to simulate cluster shutdown and recovery 

******

confirmed the fix on my node.

### DIFF
--- a/quarkchain/cluster/shard_db_operator.py
+++ b/quarkchain/cluster/shard_db_operator.py
@@ -229,26 +229,21 @@ class ShardDbOperator(TransactionHistoryMixin):
         r_minor_header_hash = r_minor_header.get_hash() if r_minor_header else b""
         self.db.put(b"r_last_m" + root_block_hash, r_minor_header_hash)
 
-    def get_root_block_by_hash(self, h, consistency_check=True):
-        """
-        Consistency check being true means whatever in memory should already have enough
-        information (no missing root block, minor block etc). Skipping the check by reading
-        directly from database may have unwanted consequences.
-        """
-        if consistency_check and h not in self.r_header_pool:
-            return None
-
+    def get_root_block_by_hash(self, h):
         raw_block = self.db.get(b"rblock_" + h, None)
         if not raw_block:
             return None
-        return RootBlock.deserialize(raw_block)
+        block = RootBlock.deserialize(raw_block)
+        self.r_header_pool[h] = block.header
+        return block
 
-    def get_root_block_header_by_hash(self, h, consistency_check=True):
+    def get_root_block_header_by_hash(self, h):
         header = self.r_header_pool.get(h, None)
-        if not header and not consistency_check:
-            block = self.get_root_block_by_hash(h, False)
+        if not header:
+            block = self.get_root_block_by_hash(h)
             if block:
                 header = block.header
+                self.r_header_pool[h] = header
         return header
 
     def get_root_block_header_by_height(self, h, height):
@@ -260,7 +255,7 @@ class ShardDbOperator(TransactionHistoryMixin):
         return r_header
 
     def contain_root_block_by_hash(self, h):
-        return h in self.r_header_pool
+        return h in self.r_header_pool or (b"rblock_" + h) in self.db
 
     def get_last_confirmed_minor_block_header_at_root_block(self, root_hash):
         """Return the latest minor block header confirmed by the root chain at the given root hash"""
--- a/quarkchain/cluster/shard_db_operator.py
+++ b/quarkchain/cluster/shard_db_operator.py
@@ -229,13 +229,27 @@ class ShardDbOperator(TransactionHistoryMixin):
         r_minor_header_hash = r_minor_header.get_hash() if r_minor_header else b""
         self.db.put(b"r_last_m" + root_block_hash, r_minor_header_hash)
 
-    def get_root_block_by_hash(self, h):
-        if h not in self.r_header_pool:
+    def get_root_block_by_hash(self, h, consistency_check=True):
+        """
+        Consistency check being true means whatever in memory should already have enough
+        information (no missing root block, minor block etc). Skipping the check by reading
+        directly from database may have unwanted consequences.
+        """
+        if consistency_check and h not in self.r_header_pool:
             return None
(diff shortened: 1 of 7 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -682,13 +682,6 @@ def run_block(
             x_shard_receive_tx_list = []
         if evm_state is None:
             evm_state = self._get_evm_state_for_new_block(block, ephemeral=False)
-        root_block_header = self.db.get_root_block_header_by_hash(
```
ninjaahhh: those two variables are not being used

ninjaahhh:
mind taking another look? @qizhou

qizhou:
working in progress.....

quarkchain/cluster/shard_db_operator.py
```diff
@@ -229,13 +229,27 @@ def put_root_block(self, root_block, r_minor_header=None, root_block_hash=None):
...
+
+        raw_block = self.db.get(b"rblock_" + h, None)
+        if not raw_block:
             return None
-        return RootBlock.deserialize(self.db.get(b"rblock_" + h))
+        return RootBlock.deserialize(raw_block)
 
-    def get_root_block_header_by_hash(self, h):
-        return self.r_header_pool.get(h, None)
+    def get_root_block_header_by_hash(self, h, consistency_check=True):
```
qizhou: contain_root_block_by_hash() should also have consistency_check field?

Issue #422, opened 2019-03-20
disable jsonrpc service by default


qizhou:
jsonrpc is not part of p2p protocol and should be used internally.  If a user wants to turn it on, the user should be responsible for the access control of the service, such as preventing it being used by public via firewall.

qizhou:
Another way is to set default host as "localhost" to prevent remote use unless explicitly specified by the user

jyouyj:
We fix this problem by disabling public and private json rpc by default value localhost. https://github.com/QuarkChain/pyquarkchain/pull/426

PR #426, opened 2019-03-22
disable_jsonrpc_by_default

jyouyj:
Should fix:  #423, #422, #416, #415, #394

We disable public and private json rpc by default value localhost to prevent it from being used by the public.  If the other people want to call it, the owner should turn it on in the configuration.

### DIFF
--- a/quarkchain/cluster/cluster_config.py
+++ b/quarkchain/cluster/cluster_config.py
@@ -135,6 +135,8 @@ class ClusterConfig(BaseConfig):
     P2P_PORT = 38291
     JSON_RPC_PORT = 38391
     PRIVATE_JSON_RPC_PORT = 38491
+    JSON_RPC_HOST = "127.0.0.1"
+    PRIVATE_JSON_RPC_HOST = "127.0.0.1"
     ENABLE_TRANSACTION_HISTORY = False
 
     DB_PATH_ROOT = "./db"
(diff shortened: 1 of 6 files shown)


quarkchain/cluster/cluster_config.py
```diff
@@ -135,6 +135,8 @@ class ClusterConfig(BaseConfig):
     P2P_PORT = 38291
     JSON_RPC_PORT = 38391
     PRIVATE_JSON_RPC_PORT = 38491
+    JSON_RPC_HOST = "127.0.0.1"
```
qizhou: 127.0.0.1 is the loopback ip for IPv4 and for IPv6, it should be  ::1.  localhost may be preferred as localhost will be automatically translated based on IPv4 or IPv6.

quarkchain/cluster/cluster_config.py
```diff
@@ -248,6 +250,14 @@ def attach_arguments(cls, parser):
             default=ClusterConfig.PRIVATE_JSON_RPC_PORT,
             type=int,
         )
+        parser.add_argument(
+            "--json_rpc_host", default=ClusterConfig.JSON_RPC_HOST, type=str
```
qizhou: Should use the same code format as private (and others)

qcdll:
I think this caused CI to break, can you please fix it @jyouyj

PR #429, opened 2019-03-22
blacklist IP for connection or handshake failures; remove blacklist from routing table

### LINKED ISSUE #427: blacklist peers
There are quite a few cases we want to blacklist a peer for certain period of time (cooldown period):
1. failed handshake
2. disconnected with `close with error`

The blacklist will be effective for both outgoing and incoming peer TCP connections.

qcdll:
fixes #427

we do not blacklist on `close_with_error` yet, I think 24-hour blacklist is too harsh for it

### DIFF
--- a/quarkchain/p2p/peer.py
+++ b/quarkchain/p2p/peer.py
@@ -70,6 +70,7 @@ from .constants import (
     HEADER_LEN,
     MAC_LEN,
     BLACKLIST_COOLDOWN_SEC,
+    UNBLACKLIST_INTERVAL,
 )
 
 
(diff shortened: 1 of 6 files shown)


quarkchain/p2p/peer.py
```diff
@@ -931,6 +936,19 @@ def _add_peer(self, peer: BasePeer, msgs: Tuple[PeerMessage, ...]) -> None:
     async def _cleanup(self) -> None:
         await self.stop_all_peers()
 
+    def blacklist(self, remote_address: Address) -> None:
+        cooldown_sec = numpy.random.exponential(BLACKLIST_COOLDOWN_SEC)
+        self.blacklist[remote_address.ip] = time_ms() // 1000 + cooldown_sec
+
+    def chk_blacklist(self, remote_address: Address) -> Bool:
+        if remote_address.ip not in self.blacklist:
```
qizhou: We probably need to check subnet rather than address (similar to Bitcoin)

quarkchain/p2p/peer.py
```diff
@@ -823,6 +825,10 @@ def __init__(
         self.listen_port = listen_port
         self.dialedout_pubkeys = set()  # type: Set[datatypes.PublicKey]
 
+        self.whitelist_nodes = whitelist_nodes
```
qizhou: whitelist_nodes or {}?

Issue #399, opened 2019-03-13
Unlimited P2P Command Size Enables Mem Crash DoS Attack


xhyumiracle:
P2P_COMMAND_SIZE_LIMIT is set to 2^32-1, which is too large for command packet. This enables attackers to send data packet as large as 4G each time to any master/slave node, which may crash the node server by occupying the memory. 
Moreover, due to the design of cluster, the whole cluster would shutdown if any master/slave subprocess exit.
```python
P2P_COMMAND_SIZE_LIMIT = (2 ** 32) - 1  # unlimited right now
```
in protocol.py:
```python
async def read_metadata_and_raw_data(self):
        ...
        if size > self.env.quark_chain_config.P2P_COMMAND_SIZE_LIMIT:
            raise RuntimeError("{}: command package exceed limit".format(self.name))
        ...
```

xhyumiracle:
BTW, here is where the read happens:
```python
raw_data_without_size = await self.__read_fully(1 + 8 + size)
```
And __read_fully keeps extending the bytearray until the size reaches n.
```python
    async def __read_fully(self, n, allow_eof=False):
        ba = bytearray()
        bs = await self.reader.read(n)
        if allow_eof and len(bs) == 0 and self.reader.at_eof():
            return None

        ba.extend(bs)
        while len(ba) < n:
            bs = await self.reader.read(n - len(ba))
            if len(bs) == 0 and self.reader.at_eof():
                raise RuntimeError("{}: read unexpected EOF".format(self.name))
            ba.extend(bs)
        return ba
```

qizhou:
This is an interesting finding.  Any suggestions from Bitcoin core or Ethereum code?  I know Bitcoin is restricting block size to be up to 1M, but glad to know if you can pinpoint the actual code is.  Thanks.

qizhou:
Bitcoin's command size is 4M (and max serialization size is 32M).  We should adopt a similar number (may be a bit larger), and limit the command size for batch download such as GET_ROOT_BLOCK_LIST_REQUEST/GET_ROOT_BLOCK_LIST_RESPONSE and GET_MINOR_BLOCK_HEADER_LIST_REQUEST/GET_MINOR_BLOCK_HEADER_LIST_RESPONSE

qizhou:
Ethereum's frame size is up to 16MB

[...]

PR #431, opened 2019-03-22
Add P2P command size limit

qizhou:
evaluate tx max size
    add more value limit checks in transactions.py
    limit P2P command size to 128M
    limit batch minor block download size to 50


qizhou:
> LGTM, what about root blocks sync? can batch root block exceed the command size limit?

Currently root block header is 249 bytes, minor block header is 479, each root block contains at most 10 blocks (may be smaller).  So to reach 128M / 100 (batch size now), we could support (1.28M - 249) / 479 / 10 = 267 shards, which is good enough for short- and long-terms.

qcdll:
fixes #399

PR #433, opened 2019-03-23
separate dialout and dialin blacklist

qcdll:
while the blacklist is effective preventing the node from dialing out to nodes that failed before (thus wasting opportunity to connect to good peers), this is a bit too harsh as we want to give the nodes a second chance to dial us

it is hard to strike a balance between security and connectivity. I believe this PR will retain security and improve connectivity

this will follow the rules defined in  https://github.com/QuarkChain/pyquarkchain/wiki/p2p-blacklisting

### DIFF
--- a/quarkchain/p2p/peer.py
+++ b/quarkchain/p2p/peer.py
@@ -69,7 +69,7 @@ from .constants import (
     DEFAULT_PEER_BOOT_TIMEOUT,
     HEADER_LEN,
     MAC_LEN,
-    BLACKLIST_COOLDOWN_SEC,
+    DIALOUT_BLACKLIST_COOLDOWN_SEC,
     UNBLACKLIST_INTERVAL,
 )
 
@@ -828,7 +828,8 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
 
         self.whitelist_nodes = whitelist_nodes or []
         # IP to unblacklist time, we blacklist by IP
-        self._blacklist = {}  # type: Dict[str, int]
+        self._dialout_blacklist = {}  # type: Dict[str, int]
+        self._dialin_blacklist = {}  # type: Dict[str, int]
 
     # async def handle_peer_count_requests(self) -> None:
     #     async def f() -> None:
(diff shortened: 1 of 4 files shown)


quarkchain/p2p/peer.py
```diff
@@ -939,31 +940,53 @@ def _add_peer(self, peer: BasePeer, msgs: Tuple[PeerMessage, ...]) -> None:
...
+            del self._dialout_blacklist[remote_address.ip]
+            return False
+        return True
+
+    def dialin_blacklist(self, remote_address: Address) -> None:
+        # never blacklist bootstap
+        for node in self.whitelist_nodes:
+            if node.address.ip == remote_address.ip:
+                return
+        self._dialin_blacklist[remote_address.ip] = time_ms() // 1000 + DIALOUT_BLACKLIST_COOLDOWN_SEC
```
qizhou: Suggest use different const name (e.g., DAILIN_BLACKLIST_COOLDOWN_SEC), and I guess they should have different values to improve connectivity :)

quarkchain/p2p/peer.py
```diff
@@ -939,31 +940,53 @@ def _add_peer(self, peer: BasePeer, msgs: Tuple[PeerMessage, ...]) -> None:
...
+            del self._dialin_blacklist[remote_address.ip]
             return False
         return True
 
     async def _periodically_unblacklist(self) -> None:
         while self.is_operational:
             now = time_ms() // 1000
             remove = set()
-            for ip, t in self._blacklist.items():
+            for ip, t in self._dialout_blacklist.items():
```
qizhou: for blacklist in (self._dailout..., self._dailin...)
   remove = []
   for ip, t in blacklist ....
   for ip in remove:

quarkchain/p2p/peer.py
qcdll: yeah right, I forgot to change this

quarkchain/p2p/peer.py
qcdll: 👍

PR #438, opened 2019-03-26
Shard DB mem only as write-through cache

ninjaahhh:
as discussed offline, can simply use in-memory `r_header_pool` as cache and remove `consistency_check` flag


ninjaahhh:
@qizhou take a look when you have time

ninjaahhh:
> Should contain_root_block_by_hash() method be also updated to check disk?

done

PR #451, opened 2019-04-01
do not serialize/deserialize 0 values in TestTokenBalanceMap

### LINKED ISSUE #405: coinbase_amount_map should ignore 0 balance tokens
in addition to disallowing 0 gas price tx, we should prevent 0 balance from coinbase_amount_map

qcdll:
fixes #405

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -177,12 +177,18 @@ class PrependedSizeListSerializer:
 
 
 class PrependedSizeMapSerializer:
+    """
+    does not serialize/deserialize biguint values that evalulate to False
+    """
+
     def __init__(self, size_bytes, key_ser, value_ser):
         self.size_bytes = size_bytes
         self.key_ser = key_ser
         self.value_ser = value_ser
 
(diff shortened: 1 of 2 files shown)


quarkchain/core.py
```diff
@@ -177,12 +177,18 @@ def deserialize(self, bb):
...
+    does not serialize/deserialize biguint values that evalulate to False
+    """
+
     def __init__(self, size_bytes, key_ser, value_ser):
         self.size_bytes = size_bytes
         self.key_ser = key_ser
         self.value_ser = value_ser
 
     def serialize(self, item_map: Dict, barray):
+        if self.value_ser is biguint:
```
qizhou: I am not a big fan of this - PrependedSizeMapSerializer should be generally purpose map serializer which should produce exactly the same map after deserialization.  I would suggest adding an argument in ctor like skip_func(k, v), and TokenBalanceMap should provide lambda k, v: v == 0 as the argument.

qcdll: sounds good, will do that, I don't like this either but didn't thought of a better way

PR #455, opened 2019-04-02
add ping/pong command and ser map

### LINKED ISSUE #452: p2p read timeout and ping/pong
replaces #401 

a few questions/up for discussion:
1. do we need a timeout for p2p read?
https://github.com/QuarkChain/pyquarkchain/blob/33e672242dea2b12745a9cbd26470d23ba18c11f/quarkchain/p2p/p2p_manager.py#L286
2. do we need ping/pong mechanism in p2p protocol? what is the benefit of having them?

qcdll:
fixes #452

### DIFF
--- a/quarkchain/cluster/p2p_commands.py
+++ b/quarkchain/cluster/p2p_commands.py
@@ -197,11 +197,22 @@ class NewBlockMinorCommand(Serializable):
         self.block = block
 
 
+class PingPongCommand(Serializable):
+    """
+    with 32B message which is undefined at the moment
+    """
+
+    FIELDS = [("message", hash256)]
+
+    def __init__(self, message):
+        self.message = message
+
+
 class CommandOp:
     HELLO = 0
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/p2p_commands.py
```diff
@@ -197,11 +197,22 @@ def __init__(self, block):
...
+    def __init__(self, message):
+        self.message = message
+
+
 class CommandOp:
     HELLO = 0
     NEW_MINOR_BLOCK_HEADER_LIST = 1
     NEW_TRANSACTION_LIST = 2
-    GET_PEER_LIST_REQUEST = 3
+    GET_PEER_LIST_REQUEST = 3  # only handled by simple_network peers
```
qizhou: what is the purpose of enabling this command?

qcdll: simple_network uses this for trivial discovery, I double-checked that this is not handled for real p2p network

PR #459, opened 2019-04-04
[part 2] Bypass root block loading consistency check in validation

ninjaahhh:
also contains cleanup on some root state methods

unlike previous PR #457, root chain is better off keeping the consistency check because it's more prone to have incomplete (inconsistent) data - we've seen in a snapshot that we have a root block in db but it lacks some of included minor blocks.  therefore using memory to keep the canonical chain seems reasonable. 

on the contrary, shard chains seem to be ok with only using memory as cache

### DIFF
--- a/quarkchain/cluster/root_state.py
+++ b/quarkchain/cluster/root_state.py
@@ -121,9 +121,9 @@ class RootDb:
         return RootBlock.deserialize(raw_block)
 
     def get_root_block_header_by_hash(self, h, consistency_check=True):
-        header = self.r_header_pool.get(h, None)
+        header = self.r_header_pool.get(h)
         if not header and not consistency_check:
-            block = self.get_root_block_by_hash(h, False)
+            block = self.get_root_block_by_hash(h, consistency_check=False)
             if block:
                 header = block.header
         return header
@@ -190,7 +190,7 @@ class RootDb:
         if key not in self.db:
             return None
         block_hash = self.db.get(key)
-        return self.get_root_block_by_hash(block_hash, False)
+        return self.get_root_block_by_hash(block_hash, consistency_check=False)
 
     # ------------------------- Minor block db operations --------------------------------
     def contain_minor_block_by_hash(self, h):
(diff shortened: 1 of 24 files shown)


quarkchain/cluster/root_state.py
```diff
@@ -121,9 +121,9 @@ def get_root_block_by_hash(self, h, consistency_check=True):
         return RootBlock.deserialize(raw_block)
 
     def get_root_block_header_by_hash(self, h, consistency_check=True):
-        header = self.r_header_pool.get(h, None)
+        header = self.r_header_pool.get(h)
```
qcdll: this would raise Exception instead?

qizhou:
I am not quite sure how this solves the problem of inconsistency - if the db doesn't have the data, how could we recover the cache data after restart?

quarkchain/cluster/root_state.py
ninjaahhh:
```
In [1]: d = {}

In [2]: d.get(1, None) is None
Out[2]: True

In [3]: d.get(1) is None
Out[3]: True
```

ninjaahhh: `d[1]` will raise exception

ninjaahhh:
@qizhou no it doesn't solve the inconsistency issue. if db is inconsistent, we can't magically recover missing data.

however, in root state, memory is used as a way to indicate "if I have the tip in memory, then I'm pretty sure other necessary information (e.g. included minor block header) are also recovered in memory". 

not 100% sure this kind of semantics is necessary, but this allows the tip-1 hack to work even under the inconsistent db state

ninjaahhh:
> I am wondering, how did you identify unused imports?

all hail pycharm

qizhou:
tip-1 trick is only used when we start a node, right?  I am still confused that since we need to recover the data from db, how cache could help?

ninjaahhh:
> tip-1 trick is only used when we start a node, right? I am still confused that since we need to recover the data from db, how cache could help?

the memory cache holds the following invariant: **blockchain info in memory should always be consistent and complete.** the `tip-1` hack simply helps keep the invariant.

the benefits: basically the full node is more pessimistic on its data. one example - 

- after restarting, say, the node has inconsistent data on an older fork, then intentionally keeping it out of memory can force the node to re-download the fork. imagine the contrary if we don't use memory to hold the canonical chain, when querying the inconsistent fork, we will have something in storage but missing something else, thus may encounter some unexpected scenarios

PR #461, opened 2019-04-04
add handler for p2p ping/pong

### LINKED ISSUE #452: p2p read timeout and ping/pong

qcdll:
fixes #452

### DIFF
--- a/quarkchain/cluster/simple_network.py
+++ b/quarkchain/cluster/simple_network.py
@@ -43,7 +43,7 @@ class Peer(P2PConnection):
             op_ser_map=OP_SERIALIZER_MAP,
             op_non_rpc_map=OP_NONRPC_MAP,
             op_rpc_map=OP_RPC_MAP,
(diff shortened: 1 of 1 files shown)


qcdll:
good point, but note that we do not use devp2p

Here is the difference: devp2p uses rplx for transport protocol whereas we have a different protocol. In this example of ping/pong command, there is actually NO command in devp2p layer (which only handles handshake TCP packets), all subsequent packets are interpreted as encrypted QKC commands. And note that timeout is also added in QKC protocol layer, so even if we have ping/pong in a layer underneath, it will not actually avoid timeouts

PR #462, opened 2019-04-04
enforce block/blockheader limit from p2p peers

### LINKED ISSUE #441: handle_get_minor_block_header_list_request中 limit 未限制导致的DOS问题
在shard.py的handle_get_minor_block_header_list_request函数中，没有对limit作上限的限制，如下
```
    async def handle_get_minor_block_header_list_request(self, request):
        if request.branch != self.shard_state.branch:
            self.close_with_error("Wrong branch from peer")
        if request.limit <= 0:
            self.close_with_error("Bad limit")
        # TODO: support tip direction
        if request.direction != Direction.GENESIS:
            self.close_with_error("Bad direction")

        block_hash = request.block_hash
        header_list = []
        for i in range(request.limit):
            header = self.shard_state.db.get_minor_block_header_by_hash(
                block_hash, consistency_check=False
            )
            header_list.append(header)
            if header.height == 0:
                break
            block_hash = header.hash_prev_minor_block

        return GetMinorBlockHeaderListResponse(
            self.shard_state.root_tip, self.shard_state.header_tip, header_list
        )
```
当传入的limit较大时，其中的header_list将不断增大，从而填充内存空间，最终可能导致内存崩溃。
而此处的request是通过p2p发送，所以我们可以作为攻击节点给目标节点发送请求攻击对应shard的slave服务器，此处我也在本地做了简单的尝试，不过因为本地并没有足够的区块，所以便没有每次向前取区块，真正攻击可能需要网络上线几个月后有足够的区块才行。
当发送limit为100万左右时，大概一分钟后内存占用如下，而且对应的处理线程会卡住

内存占用增长的很快，如果我们进一步增大limit的话，后果就更可怕了

造成的内存压力很大

建议对 request 中的 limit 设置上限。

qcdll:
fixes #441 
one major issue that was also fixed in this PR is bad peer could potentially cause cluster crash (because we `check` that returned number of blocks matches requested, which is totally controlled by peer):
https://github.com/QuarkChain/pyquarkchain/blob/6a66a11ad48835163a20b3b06ed3ff59a35dde11/quarkchain/cluster/shard.py#L307

maybe we should have some audit on those unsafe `check`s as well

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -31,6 +31,13 @@ from quarkchain.utils import Logger, check, time_ms
 from quarkchain.p2p.utils import RESERVED_CLUSTER_PEER_ID
 
 
+SYNC_TIMEOUT = 10
+# Current minor block size is up to 6M gas / 4 (zero-byte gas) = 1.5M
+# Per-command size is now 128M so 128M / 1.5M = 85
+MINOR_BLOCK_BATCH_SIZE = 50
+MINOR_BLOCK_HEADER_LIST_LIMIT = 100
+
+
 class PeerShardConnection(VirtualConnection):
     """ A virtual connection between local shard and remote shard
     """
@@ -99,7 +106,7 @@ class PeerShardConnection(VirtualConnection):
     async def handle_get_minor_block_header_list_request(self, request):
         if request.branch != self.shard_state.branch:
             self.close_with_error("Wrong branch from peer")
-        if request.limit <= 0:
+        if request.limit <= 0 or request.limit > 2 * MINOR_BLOCK_HEADER_LIST_LIMIT:
             self.close_with_error("Bad limit")
         # TODO: support tip direction
         if request.direction != Direction.GENESIS:
@@ -121,6 +128,8 @@ class PeerShardConnection(VirtualConnection):
         )
 
     async def handle_get_minor_block_list_request(self, request):
+        if len(request.minor_block_hash_list) > 2 * MINOR_BLOCK_BATCH_SIZE:
+            self.close_with_error("Bad number of minor blocks requested")
         m_block_list = []
         for m_block_hash in request.minor_block_hash_list:
             m_block = self.shard_state.db.get_minor_block_by_hash(
@@ -216,11 +225,6 @@ OP_RPC_MAP = {
     ),
 }
 
-TIMEOUT = 10
-# Current minor block size is up to 6M gas / 4 (zero-byte gas) = 1.5M
-# Per-command size is now 128M so 128M / 1.5M = 85
-MINOR_BLOCK_BATCH_SIZE = 50
-
 
 class SyncTask:
     """ Given a header and a shard connection, the synchronizer will synchronize
@@ -275,7 +279,7 @@ class SyncTask:
                 )
             )
             block_header_list = await asyncio.wait_for(
-                self.__download_block_headers(block_hash), TIMEOUT
+                self.__download_block_headers(block_hash), SYNC_TIMEOUT
             )
             Logger.info(
                 "[{}] downloaded {} headers from peer".format(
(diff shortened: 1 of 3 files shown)


quarkchain/cluster/shard.py
```diff
@@ -297,14 +301,18 @@ def __init__(self, header: MinorBlockHeader, shard_conn: PeerShardConnection):
...
             )
             Logger.info(
                 "[{}] downloaded {} blocks from peer".format(
                     self.shard_state.branch.to_str(), len(block_chain)
                 )
             )
-            check(len(block_chain) == len(block_header_chain[:MINOR_BLOCK_BATCH_SIZE]))
+            if len(block_chain) != len(block_header_chain[:MINOR_BLOCK_BATCH_SIZE]):
+                # TODO: tag bad peer
+                return self.shard_conn.close_with_error(
```
ninjaahhh: yeah agree this is way better than `check`

qcdll:
yeah, we currently have the test_cluster which does have p2p communications, just need to add coverage over there

quarkchain/cluster/simple_network.py
```diff
@@ -25,6 +25,10 @@
 from quarkchain.utils import Logger
 
 
+ROOT_BLOCK_BATCH_SIZE = 100
+ROOT_BLOCK_HEADER_LIST_LIMIT = 500
```
qizhou: Could we find any reference number from Ethereum?  Looks like go-ethereum uses 128 and 512 respectively.

https://github.com/ethereum/go-ethereum/blob/master/eth/downloader/downloader.go

quarkchain/cluster/shard.py
qizhou: I think the protocol should be
- the peer could response a smaller amount of blocks since it may implement its local response limit
- The responded blocks should maintain the same order as requested
- The responded blocks should be at least 1

quarkchain/cluster/simple_network.py
```diff
@@ -228,7 +232,7 @@ def get_connection_to_forward(self, metadata):
             await self.master_server.add_transaction(tx, self)
 
     async def handle_get_root_block_header_list_request(self, request):
-        if request.limit <= 0:
+        if request.limit <= 0 or request.limit > 2 * ROOT_BLOCK_HEADER_LIST_LIMIT:
```
qizhou: I think request.limit should be min(request.limit, ROOT_BLOCK_HEADER_LIST_LIMIT).  This means ROOT_BLOCK_HEADER_LIST_LIMIT won't be a parameter in consensus.

quarkchain/cluster/simple_network.py
```diff
@@ -247,6 +251,8 @@ def get_connection_to_forward(self, metadata):
         return GetRootBlockHeaderListResponse(self.root_state.tip, header_list)
 
     async def handle_get_root_block_list_request(self, request):
+        if len(request.root_block_hash_list) > 2 * ROOT_BLOCK_BATCH_SIZE:
```
qizhou: See above.

quarkchain/cluster/simple_network.py
qcdll: we don't want ROOT_BLOCK_HEADER_LIST_LIMIT to be part of consensus, just a way to avoid DOS from evil peers

quarkchain/cluster/shard.py
qcdll: I think there is freedom in how this is handled. The remote peer is free to respond with any number of blocks they may want to return (this PR fixes the potential risk that an evil peer will crash this node) and this node is free to disconnect if the number does not meet expectations

qcdll:
The proposals sound fine but we want to fix potential security hazard first and improve protocol for block download in later releases

ninjaahhh:
how about making an issue to better track the aforementioned protocol improvement?

qizhou:
like QCIP (QuarkChain Improvement Proposal)?

 - Qi

PR #464, opened 2019-04-05
New block pool#387

jyouyj:
Check difficulty number is correct according to parent difficulty

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -69,8 +69,8 @@ class PeerShardConnection(VirtualConnection):
     def broadcast_new_tip(self):
         if self.best_root_block_header_observed:
             if (
-                self.shard_state.root_tip.height
-                < self.best_root_block_header_observed.height
+                self.shard_state.root_tip.total_difficulty
+                < self.best_root_block_header_observed.total_difficulty
             ):
                 return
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/shard.py
```diff
@@ -579,6 +579,36 @@ def broadcast_tx_list(self, tx_list, source_peer=None):
             ]
             consensus_type = shard_config.CONSENSUS_TYPE
             diff = header.difficulty
+
+            # Check difficulty
+            if not self.env.quark_chain_config.SKIP_MINOR_DIFFICULTY_CHECK:
+                if not self.state.db.contain_minor_block_by_hash(
```
qcdll: this requires the prev block to be in DB, but there may be cases where the parent may also still be in the new block pool. This doesn't look right

quarkchain/cluster/shard.py
```diff
@@ -76,8 +76,8 @@ def send_new_block(self, block):
     def broadcast_new_tip(self):
         if self.best_root_block_header_observed:
             if (
-                self.shard_state.root_tip.height
-                < self.best_root_block_header_observed.height
+                self.shard_state.root_tip.total_difficulty
```
ninjaahhh: please merge with master to show the correct and relevant changes

jyouyj:
I will submit a new pull request later.

PR #469, opened 2019-04-10
Add more strict check on new block pool

### LINKED ISSUE #387: new_block_pool 未及时清理，可能会引发DOS攻击
注意到在shard.py中未对new_block_pool做及时的清理，在add_block函数中的逻辑如下
```
        try:
            xshard_list, coinbase_amount_map = self.state.add_block(block)
        except Exception as e:
            Logger.error_exception()
            return False

        # only remove from pool if the block successfully added to state,
        #   this may cache failed blocks but prevents them being broadcasted more than needed
        # TODO add ttl to blocks in new_block_pool
        self.state.new_block_pool.pop(block.header.get_hash(), None)
```
在state中add_block若失败则直接返回，new_block_pool中的失败区块却没有进行清理，这样它们将持续堆积在内存中。
从攻击者角度上讲我们可以构造大量的恶意区块，它们无法通过state.add_block中的验证，但却会被添加进new_block_pool中，对节点的内存进行DOS攻击。不过在添加进new_block_pool之前会对该区块的工作量进行检查，这样就需要我们耗费一定的资源，但是因为在接收区块时并未对高度进行限制，那么我们可以从区块高度非常低时在同一高度生成大量区块，此时需要的难度较低，而且这样正常流程生成的区块本身也会在state.add_block中失败，满足了我们的要求。另外此处的validate_seal函数似乎并未对diff进行检查，只是一个简单估计，应该可以直接将diff调整到最小值，这样也可以生成大量恶意区块。利用这些区块即可对目标节点的内存进行DOS攻击。

jyouyj:
Fix #387 and check difficulty number is correct according to parent difficulty

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -573,6 +573,7 @@ class Shard:
             > time_ms() // 1000 + ALLOWED_FUTURE_BLOCKS_TIME_BROADCAST
         ):
             return
+        # Ignore old blocks
         if (
             self.state.header_tip
             and self.state.header_tip.height - block.header.height
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -28,12 +28,10 @@ from quarkchain.core import (
 )
 from quarkchain.constants import (
     ALLOWED_FUTURE_BLOCKS_TIME_BROADCAST,
-    ALLOWED_FUTURE_BLOCKS_TIME_VALIDATION,
     NEW_TRANSACTION_LIST_LIMIT,
     MINOR_BLOCK_BATCH_SIZE,
     MINOR_BLOCK_HEADER_LIST_LIMIT,
     SYNC_TIMEOUT,
-    ALLOWED_STALE_BLOCKS_HEIGHT_DIFF,
 )
 from quarkchain.db import InMemoryDb, PersistentDb
 from quarkchain.utils import Logger, check, time_ms
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -33,6 +33,7 @@ from quarkchain.constants import (
     MINOR_BLOCK_BATCH_SIZE,
     MINOR_BLOCK_HEADER_LIST_LIMIT,
     SYNC_TIMEOUT,
+    ALLOWED_STALE_BLOCKS_HEIGHT_DIFF,
 )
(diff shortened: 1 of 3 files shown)


ninjaahhh:
to address all the problem mentioned in #387, following changes also seem necessary

1. in `shard.add_block()`, when `self.state.add_block(block)` failed, the block should be popped from `new_block_pool`
2. limit how far away `handle_new_block` should handle. otherwise we still will receive new blocks with very low diff (block height 3)

also, what do you think if we move the full validation logic to `shard.handle_new_block` instead of `shard_state.add_block`?

qcdll:
I agree that this only fixes a portion of #387 

to me, `handle_new_block` is a way to handle broadcasted blocks and we need to decide whether to accept and re-broadcast; `shard_state.add_block` is the real deal and goes through the most stringent checks. I feel a proper way may be have a preliminary check enforced by `handle_new_block` and have `shard_state.add_block` run it as a subroutine as well

ninjaahhh:
realized we intentionally keep the bad block in `new_block_pool` to avoid spamming others, as mentioned in the comments. the ideal fix would be to add a TTL for items in the pool

for now will keep the bad block in the pool

quarkchain/cluster/shard.py
```diff
@@ -560,15 +559,30 @@ def broadcast_tx_list(self, tx_list, source_peer=None):
...
+
+        # Sanity check on timestamp and block height
+        if (
+            block.header.create_time
+            > time_ms() // 1000 + ALLOWED_FUTURE_BLOCKS_TIME_BROADCAST
         ):
-            if block.header.hash_prev_minor_block not in self.state.new_block_pool:
-                return
+            return
+        if (
```
qizhou: May add a comment
# Skip the block if it is too old.

quarkchain/cluster/shard.py
```diff
@@ -560,15 +559,30 @@ def broadcast_tx_list(self, tx_list, source_peer=None):
...
+            self.state.header_tip
+            and self.state.header_tip.height - block.header.height
+            > self.state.shard_config.max_stale_minor_block_height_diff
+        ):
+            return
```
ninjaahhh: it's essentially the same check when adding the block

https://github.com/QuarkChain/pyquarkchain/blob/178d8976a26989e13bd1aee0c9dfb6221cb2260a/quarkchain/cluster/shard_state.py#L812-L841

just add it here as a prerequisite

PR #471, opened 2019-04-10
Update xshard deposit format to indicate coinbase

### LINKED ISSUE #470: Weird tx are indexed for miner
example

http://testnet2.quarkchain.io/address/0xd50f23e410711c1391f5ef88fc11245e564c768400000000

I guess those payments are block rewards but somehow regarded as tx

UPDATE: they are all root block rewards, which are represented as x-shard deposits

ninjaahhh:
in #350, [it  changes the format of x-shard deposit](https://github.com/QuarkChain/pyquarkchain/pull/350/files#diff-87764b6e3711103099757595c6f871c1L1242), which breaks the indexer's assumption (root block coinbase reward deposit should have 0 tx hash):

https://github.com/QuarkChain/pyquarkchain/blob/c89084205605a9dd505008434b7098790a2a1733/quarkchain/cluster/shard_db_operator.py#L69-L75

this PR added `is_coinbase` field to xshard deposit so client can handle correspondingly. fixes #470

### DIFF
--- a/quarkchain/cluster/jsonrpc.py
+++ b/quarkchain/cluster/jsonrpc.py
@@ -839,6 +839,7 @@ class JSONRPCServer:
                     "blockHeight": quantity_encoder(tx.block_height),
                     "timestamp": quantity_encoder(tx.timestamp),
                     "success": tx.success,
+                    "is_coinbase": tx.is_coinbase,
                 }
             )
         return {"txList": txs, "next": data_encoder(next)}
--- a/quarkchain/cluster/jsonrpc.py
+++ b/quarkchain/cluster/jsonrpc.py
@@ -393,10 +393,10 @@ class JSONRPCServer:
     def start_public_server(cls, env, master_server):
         server = cls(
             env,
-            master_server, 
+            master_server,
             env.cluster_config.JSON_RPC_PORT,
             env.cluster_config.JSON_RPC_HOST,
-            public_methods
+            public_methods,
         )
         server.start()
         return server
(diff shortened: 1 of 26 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -106,8 +106,8 @@ def __get_current_tx(self):
 
                 # Perform x-shard from root chain coinbase
                 return CrossShardTransactionDeposit(
-                    tx_hash=self.rblock.header.get_hash(),
-                    from_address=self.rblock.header.coinbase_address,
+                    tx_hash=bytes(32),
```
qizhou: May be we should use sha3(rootblock.get_hash() + "coinbase")?  This will make sure each tx_hash is unique.

ninjaahhh: any benefit for making it unique? because it will again break the indexer, which by default regards `tx_hash == bytes(32)` as root block coinbase reward

ninjaahhh: the reason for my change is to find a way to differentiate between real xshard deposit and root block coinbase, because the indexing server should only index the former

qizhou: Yes, but in general, one txhash = one tx, while now we have one txhash = a list of txs.  For index server, do you think we can resort other way to detect (but probably still need to show the coinbase tx properly).

ninjaahhh:
a clean fix would be to use another field to indicate root block coinbase reward

ninjaahhh:
@qizhou take another look when you have time

quarkchain/cluster/jsonrpc.py
```diff
@@ -839,6 +839,7 @@ def get_data_default(key, decoder, default=None):
                     "blockHeight": quantity_encoder(tx.block_height),
                     "timestamp": quantity_encoder(tx.timestamp),
                     "success": tx.success,
+                    "is_coinbase": tx.is_coinbase,
```
qizhou: May use the name "is_from_root_chain"?  Since in the future, we plan to support tx in root_chain and allow root_chain to send a transaction to different shards.

ninjaahhh: sounds good. will update the name

PR #473, opened 2019-04-10
[BREAKING CHANGE] Modify full shard key rlp in evm tx

ninjaahhh:
brought up by @pingke, seems it doesn't give much benefit using different serialization for shard key

note this is a breaking change. shouldn't merge in current network. should be for root chain.


qizhou:
Current minimum size transaction should adopt most fields in 0 (note that the transactions should be valid).  This will help us identify the max block size of a shard.

qcdll:
so it would be best to change the min size accordingly as `from_full_shard_key` can be 0

qizhou:
> so it would be best to change the min size accordingly as `from_full_shard_key` can be 0

Yes, in theory, a valid transaction could.

ninjaahhh:
close this since we want to avoid breaking changes. let's keep `BigEndianInt(4))`

ninjaahhh:
close this since we want to avoid breaking changes. let's keep `BigEndianInt(4))`

> close this since we want to avoid breaking changes. let's keep `BigEndianInt(4))`

cc @pingke

PR #475, opened 2019-04-10
tx_whitelist_senders in qkc config

### LINKED ISSUE #449: allow whitelisted address to send transactions while ENABLE_TX_TIMESTAMP is not reached
We will allow some specific sender to transfer token so that
1, distribute some bounty (such as mining) early
2, allow early token swap for PoSW users
3, etc

qcdll:
fixes #449 
move checks to proper places to check sender address

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -452,6 +452,16 @@ class ShardState:
         if evm_tx.is_cross_shard and evm_tx.startgas > xshard_gas_limit:
             raise RuntimeError("xshard evm tx exceeds xshard gas limit")
 
+        # check if TX is disabled
+        if (
+            self.env.quark_chain_config.ENABLE_TX_TIMESTAMP is not None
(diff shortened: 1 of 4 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -636,14 +646,6 @@ def __validate_block(
                 % (xshard_gas_limit, block.meta.evm_xshard_gas_limit)
             )
 
-        if (
```
ninjaahhh: I think we still need to make sure blocks only contain whitelisted tx.

note that other miners can always manually craft a block including unwanted transactions.

qcdll: then the block would be rejected as transactions cannot be validated

qcdll: this is also covered by unittest:
https://github.com/QuarkChain/pyquarkchain/pull/475/files#diff-f0cb0611ad35e203a2117908d4a5a812R2241
even the block contains transactions (unwhitelisted), it would be rejected

PR #483, opened 2019-04-13
Always return shard block's coinbase during slave sync

qizhou:
fix a root block synchronizer bug that
- a shard block is added in the local db
- the shard block is not in master (due to shutdown or sync fail)
so that it returns None coinbase_map, and not added in master db (in sync block logic)

- log root block header download time
- make shard tip receive log less verbose

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -192,10 +192,11 @@ class PeerShardConnection(VirtualConnection):
         if self.shard_state.header_tip.height >= m_header.height:
             return
 
-        Logger.info(
+        Logger.info_every_sec(
             "[{}] received new tip with height {}".format(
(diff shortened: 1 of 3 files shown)


quarkchain/cluster/shard.py
```diff
@@ -685,7 +686,9 @@ def broadcast_tx_list(self, tx_list, source_peer=None):
                 xshard_list, coinbase_amount_map = self.state.add_block(
                     block, skip_if_too_old=False
                 )
-                coinbase_amount_list.append(coinbase_amount_map)
+                # coinbase_amount_map may be None if the block exists
+                # adding the block header one since the block is already validated.
+                coinbase_amount_list.append(block.header.coinbase_amount_map)
```
ninjaahhh: also update comment at line 670

PR #489, opened 2019-04-15
Revert "Check failed transaction gas"

ninjaahhh:
since this is a breaking change, need to revert this for a new release for testnet.

will prepare another revert PR asap


qizhou:
Could we revert it in different branch such as testnet2.5.5?

ninjaahhh:
> Could we revert it in different branch such as testnet2.5.5?

yeah creating a new branch will work, my concern is in this way the code base became incompatible with the master, which means the user can't simply `git pull origin master` after downloading the release.

but agree it's not worth putting to the master history

PR #497, opened 2019-04-16
add get root block header list with skip p2p command

qizhou:
Similar to https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getblockheaders-0x03, add the command to support skip so that we could fast locate the ancestor of the local tip and remote tip in the canonical chain.

### DIFF
--- a/quarkchain/cluster/p2p_commands.py
+++ b/quarkchain/cluster/p2p_commands.py
@@ -11,6 +11,7 @@ from quarkchain.core import (
 )
 from quarkchain.core import RootBlockHeader, MinorBlockHeader, RootBlock, MinorBlock
 from quarkchain.core import Serializable, PrependedSizeListSerializer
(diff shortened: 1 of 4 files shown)


qizhou:
Currently synchronizer is too slow - taking too much time in download block headers.  The planned optimization is to locate their ancestor asap and the download from the ancestor - this requires the new command so that we could arbitrarily query a block header with its height in the canonical root chain.

qizhou:
> while this sounds fine, I don't get why downloading 500 headers at a time can cause slowness, to download 1 week of root block headers we just need to download 20 times which looks OK to me

Currently, it takes 1-10 seconds to download 500 headers, which means for 44000 blocks , it takes 88s to 880s, which is a considerable time.  in addition, any sync error will cause re-download the blocks.  With such change, we only need log(time) to start downloading the actual blocks.

PR #500, opened 2019-04-16
Root chain reorg will remove blocks in old fork with high height

qizhou:


quarkchain/cluster/root_state.py
```diff
@@ -539,8 +545,13 @@ def validate_block(self, block, block_hash=None):
 
         return block_hash, prev_header_map.values()
 
-    def __rewrite_block_index_to(self, block):
+    def __rewrite_block_index_to(self, old_block_header, new_block):
         """ Find the common ancestor in the current chain and rewrite index till block """
+        # If old block height is greater than new block's, remove the indices
+        for i in range(new_block.header.height + 1, old_block_header.height + 1):
+            self.db.remove_root_block_index(i)
```
ninjaahhh: what's the consequence of keeping those higher height blocks in db? do we have any logic that depends on this?

qizhou: The new p2p command allows downloading headers from the canonical chain by height, where the previous code allows downloading the headers with a height greater than that of the tip.

ninjaahhh: but I see you already made sure the block height <= self.root_state.tip in the new p2p command:

https://github.com/QuarkChain/pyquarkchain/blob/cf74767dfc353613be5c67d330e8ead6ce06c3c9/quarkchain/cluster/simple_network.py#L290-L302

not sure whether it's worth removing old blocks, if we can handle it in application layer easily

qizhou: Looks like the code works better than expected.  However, I personally don't want to introduce exception on the canonical chain where a block with higher height than tip could exist, i.e., anyone should easily pick up the code with common assumption without knowing the details of exceptions.  In addition, such code will rarely happen so for performance-wise, we don't need to concern.

ninjaahhh: ok sounds reasonable.

PR #502, opened 2019-04-17
ShardState.add_root_block should update tip's evm_state

qizhou:
…rrectly

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -940,15 +940,13 @@ class ShardState:
                 update_tip = prev_root_header.height > tip_prev_root_header.height
 
         if update_tip:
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -1413,11 +1419,7 @@ def add_root_block(self, root_block: RootBlock):
             )
 
         if self.header_tip != orig_header_tip:
-            header_tip_hash = self.header_tip.get_hash()
```
ninjaahhh: so in the old code, we didn't update the evm state correctly after adding the root block？

qizhou: That is correct, and the added test will capture that.

PR #504, opened 2019-04-18
add trie test cmp with go

qizhou:


qizhou:
> do we want go code in the repo? also it looks the test_case_# files are generated by the go file

Yes, the test cases are generated by the go file.  This want to make sure that both python and go trie implementations have exactly the same output.

qcdll:
yes, but I don't think we should put generated content in source repo, they can be generated on demand when running the python test

qizhou:
> yes, but I don't think we should put generated content in source repo, they can be generated on demand when running the python test

I want to include the test in the CI, however, to generate the test cases, a user must also download the go-ethereum and install it properly in GOPATH.  So I think that putting a generated test cases together with the code to generate should be a good tradeoff.  What do you think?

PR #505, opened 2019-04-18
Support guardian signature

jyouyj:
Could we consider sign the root block using get_hash_for_mining() function?
    def get_hash_for_mining(self):
        return sha3_256(self.serialize_without(["nonce", "mixhash", "signature"]))

The pool server only obtains the information of root block header through getWork and getRootBlockByHeight RPCs. Thus, it cannot obtain the full information of the root block header.

### DIFF
--- a/quarkchain/cluster/jsonrpc.py
+++ b/quarkchain/cluster/jsonrpc.py
@@ -134,6 +134,12 @@ def hash_decoder(hex_str):
         raise InvalidParams("Hashes must be 32 bytes long")
     return decoded
 
+def signature_decoder(hex_str):
+    """Decode a block signature."""
+    decoded = data_decoder(hex_str)
+    if len(decoded) != 65:
+        raise InvalidParams("Signature must be 65 bytes long")
+    return decoded
 
 def bool_decoder(data):
     if not isinstance(data, bool):
@@ -865,7 +871,8 @@ class JSONRPCServer:
     @decode_arg("header_hash", hash_decoder)
     @decode_arg("nonce", quantity_decoder)
     @decode_arg("mixhash", hash_decoder)
-    async def submitWork(self, full_shard_key, header_hash, nonce, mixhash):
+    @decode_arg("signature", signature_decoder)
+    async def submitWork(self, full_shard_key, header_hash, nonce, mixhash, signature="0x" + bytes(65).hex()):
         branch = None  # `None` means getting work from root chain
         if full_shard_key is not None:
             branch = Branch(
@@ -873,7 +880,7 @@ class JSONRPCServer:
                     full_shard_key
                 )
             )
-        return await self.master.submit_work(branch, header_hash, nonce, mixhash)
+        return await self.master.submit_work(branch, header_hash, nonce, mixhash, signature)
 
     @public_methods.add
     @decode_arg("full_shard_key", shard_id_decoder)
--- a/quarkchain/cluster/jsonrpc.py
+++ b/quarkchain/cluster/jsonrpc.py
@@ -134,12 +134,16 @@ def hash_decoder(hex_str):
         raise InvalidParams("Hashes must be 32 bytes long")
     return decoded
 
+
 def signature_decoder(hex_str):
     """Decode a block signature."""
+    if not hex_str:
+        return None
     decoded = data_decoder(hex_str)
     if len(decoded) != 65:
         raise InvalidParams("Signature must be 65 bytes long")
     return decoded
+
 
 def bool_decoder(data):
     if not isinstance(data, bool):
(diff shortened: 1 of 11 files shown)


ninjaahhh:
this is consensus change, but should help miner side to sign their work **without** the need to get the full block header.

qizhou:
I am thinking the hash should include "nonce" and "mixhash", right?  Looks like without "nonce" and "mixhash", an attack may mine another block by re-using the signature.  However, the following blocks cannot point to the new block since its hash is different than previous, so I think it is fine.

quarkchain/cluster/miner.py
```diff
@@ -281,7 +279,13 @@ def _mine_new_block_async(self):
             copy.deepcopy(self.current_work),
         )
 
-    async def submit_work(self, header_hash: bytes, nonce: int, mixhash: bytes, signature: bytes = bytes(65),) -> bool:
+    async def submit_work(
+        self,
+        header_hash: bytes,
+        nonce: int,
+        mixhash: bytes,
+        signature: Optional[bytes] = bytes(65),
```
ninjaahhh: since it's `Optional`, default arg should be `None`

```python
signature: Optional[bytes] = None
```

quarkchain/cluster/miner.py
```diff
@@ -296,12 +300,8 @@ def _mine_new_block_async(self):
             header.sign_with_private_key(self.guardian_private_key)
 
         # remote sign as a guardian
-        if isinstance(block, RootBlock) and header.signature == bytes(65) and signature != bytes(65):
+        if isinstance(block, RootBlock) and signature != bytes(65):
```
ninjaahhh: and here should check against `None`

quarkchain/cluster/master.py
```diff
@@ -562,7 +562,12 @@ def close_with_error(self, error):
         )
 
     async def submit_work(
-        self, branch: Branch, header_hash: bytes, nonce: int, mixhash: bytes, signature: bytes = bytes(65),
+        self,
+        branch: Branch,
+        header_hash: bytes,
+        nonce: int,
+        mixhash: bytes,
+        signature: bytes = bytes(65),
```
ninjaahhh: `signature: Optional[bytes] = None`

quarkchain/cluster/master.py
```diff
@@ -1357,10 +1361,17 @@ def is_mining(self):
         return await slave.get_work(branch)
 
     async def submit_work(
-        self, branch: Optional[Branch], header_hash: bytes, nonce: int, mixhash: bytes, signature: bytes = bytes(65),
+        self,
+        branch: Optional[Branch],
+        header_hash: bytes,
+        nonce: int,
+        mixhash: bytes,
+        signature: bytes = bytes(65),
```
ninjaahhh: `signature: Optional[bytes] = None`

quarkchain/core.py
```diff
@@ -909,18 +909,21 @@ def __init__(
         self.signature = signature
 
     def get_hash(self):
```
qizhou: I guess you don't need the function since get_hash() from the parent will do the same thing.

quarkchain/cluster/tests/test_miner.py
```diff
@@ -247,6 +247,50 @@ def test_submit_work_with_guardian(self):
...
+                # mimic the sign process of the remote guardian server
+                block.header.nonce = i
+                block.header.mixhash = sha3_256(b"")
+                block.header.sign_with_private_key(priv)
+                signature = block.header.signature
+
+                # reset the signature to the default value
+                block.header.signature = bytes(65)
+                # submit the signature through the submit work
+                res = await miner.submit_work(work.hash, i, sha3_256(b""), signature)
```
ninjaahhh: add another case that failed?

```python
await miner.submit_work(work.hash, i, sha3_256(b""), bytes(65))
self.assertFalse(res)
```

quarkchain/cluster/jsonrpc.py
```diff
@@ -134,6 +134,12 @@ def hash_decoder(hex_str):
         raise InvalidParams("Hashes must be 32 bytes long")
     return decoded
 
+def signature_decoder(hex_str):
+    """Decode a block signature."""
+    decoded = data_decoder(hex_str)
```
ninjaahhh:
```
if not hex_str:
  return None
```

quarkchain/cluster/jsonrpc.py
```diff
@@ -865,15 +871,16 @@ def get_data_default(key, decoder, default=None):
     @decode_arg("header_hash", hash_decoder)
     @decode_arg("nonce", quantity_decoder)
     @decode_arg("mixhash", hash_decoder)
-    async def submitWork(self, full_shard_key, header_hash, nonce, mixhash):
+    @decode_arg("signature", signature_decoder)
+    async def submitWork(self, full_shard_key, header_hash, nonce, mixhash, signature="0x" + bytes(65).hex()):
```
ninjaahhh: default arg `signature=None`

PR #510, opened 2019-04-19
Add committing state of root block to handle re-broadcast

ninjaahhh:
fixing tests. idea is simple, only commit a root block if it successfully broadcasted to slaves

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -1036,8 +1036,6 @@ class MasterServer:
         """ Add root block locally and broadcast root block to all shards and .
         All update root block should be done in serial to avoid inconsistent global root block state.
         """
-        self.root_state.validate_block(r_block)  # throw exception if failed
-
         # use write-ahead log so if crashed the root block can be re-broadcasted
         self.root_state.write_committing_hash(r_block.header.get_hash())
 
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -689,6 +689,10 @@ class MasterServer:
             r_block = self.root_state.db.get_root_block_by_hash(
                 committing_block_hash, consistency_check=False
             )
+            # missing actual block, may have crashed before writing the block
+            if not r_block:
+                self.root_state.clear_committing_hash()
+                return
             future_list = self.broadcast_rpc(
                 op=ClusterOp.ADD_ROOT_BLOCK_REQUEST,
                 req=AddRootBlockRequest(r_block, False),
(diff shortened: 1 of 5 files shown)


quarkchain/cluster/master.py
```diff
@@ -1016,6 +1017,11 @@ def handle_new_root_block_header(self, header, peer):
         """ Add root block locally and broadcast root block to all shards and .
         All update root block should be done in serial to avoid inconsistent global root block state.
         """
+        block_hash = r_block.header.get_hash()
+        committed = self.root_state.is_committed_by_hash(block_hash)
```
qizhou: => if self.root_state.is_committed_by_hash(...)  so that we could save one variable?

quarkchain/cluster/master.py
```diff
@@ -1017,13 +1033,15 @@ def handle_new_root_block_header(self, header, peer):
...
             update_tip = self.root_state.add_block(r_block)
-            success = True
         except ValueError:
             Logger.log_exception()
-            success = False
+            return
+
+        # root block is written to db but not broadcasted to slaves
+        # use write-ahead log so if crashed the root block can be re-broadcasted
+        self.root_state.write_committing_hash(r_block.header.get_hash())
```
qizhou: I guess we need to put the code before add_block - if shutdown happens between add_block and write_committing_hash, and restarting the node won't recover the correct block

quarkchain/cluster/master.py
```diff
@@ -1017,13 +1033,15 @@ def handle_new_root_block_header(self, header, peer):
         All update root block should be done in serial to avoid inconsistent global root block state.
         """
         self.root_state.validate_block(r_block)  # throw exception if failed
```
qizhou: Looks like validate_block could be removed (since add_block will do the same thing)?

quarkchain/cluster/master.py
```diff
@@ -682,6 +683,20 @@ def __get_mining_params():
             guardian_private_key=self.env.quark_chain_config.guardian_private_key,
         )
 
+    async def __rebroadcast_committing_root_block(self):
+        committing_block_hash = self.root_state.get_committing_block_hash()
+        if committing_block_hash:
+            r_block = self.root_state.db.get_root_block_by_hash(
+                committing_block_hash, consistency_check=False
+            )
+            future_list = self.broadcast_rpc(
```
qizhou: If we move committing_block_hash() before add_block() (see below), r_block may be None and we should just skip re-broadcasting.  If r_block is not None, a log message may be preferred.

qizhou:
oh, once race is that add_root_block may be called by local miner, which will allow two root block add in-flight......

quarkchain/cluster/master.py
ninjaahhh: on the other hand, after changing the order, if it crashes between writing committing hash and the root block, we won't find the root block pointed by the committing hash - which should be fine since it's not broadcasted yet, thus can safely ignore.

good find

ninjaahhh:
take another look when you have time @qizhou

quarkchain/cluster/master.py
```diff
@@ -1041,14 +1061,14 @@ def handle_new_root_block_header(self, header, peer):
...
+        # use write-ahead log so if crashed the root block can be re-broadcasted
+        self.root_state.write_committing_hash(r_block.header.get_hash())
+
         try:
             update_tip = self.root_state.add_block(r_block)
-            success = True
         except ValueError:
             Logger.log_exception()
-            success = False
+            return
```
qizhou: should re-throw?

ninjaahhh: ok

ninjaahhh:
> Looks like to me. But be aware that is a window that two add_root_blocks are in-flight
> 
> * The synchronizer is downloading the block
> * Local miner is submitting the block
> 
> We may need to address the race later.

we should track this issue in the wiki. can you add it?

PR #511, opened 2019-04-19
optimize root block synchronizer

qizhou:
Using n-ary method to lookup ancestor of local and remote chains with the following benefits:
- Should take about at most 2 RPCs to locate the best ancestor if we download 500 headers per RPC.  The first RPC will divide at most 22500 blocks to 499 bins with each bin has at most 46 elements.  The next RPC could local the ancestor directly.
- The code assumes the remote chain may be organized significantly.  Special handling is required (currently we just abort the sync)
- After locating ancestor, we will download the blocks from ancestor to the remote tip, and make incremental sync progress.

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -52,6 +52,7 @@ from quarkchain.cluster.rpc import (
     SubmitWorkRequest,
     SubmitWorkResponse,
     AddMinorBlockHeaderListResponse,
+    RootBlockSychronizerStats,
 )
 from quarkchain.cluster.rpc import (
     ConnectToSlavesRequest,
@@ -92,7 +93,7 @@ class SyncTask:
     including root chain and shards with the peer up to the height of the header.
     """
 
-    def __init__(self, header, peer):
+    def __init__(self, header, peer, stats, root_block_header_list_limit):
         self.header = header
         self.peer = peer
         self.master_server = peer.master_server
@@ -100,6 +101,9 @@ class SyncTask:
         self.max_staleness = (
             self.root_state.env.quark_chain_config.ROOT.MAX_STALE_ROOT_BLOCK_HEIGHT_DIFF
         )
+        self.stats = stats
+        self.root_block_header_list_limit = root_block_header_list_limit
+        check(root_block_header_list_limit >= 3)
 
     async def sync(self):
         try:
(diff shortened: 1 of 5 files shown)


quarkchain/cluster/master.py
```diff
@@ -107,119 +108,127 @@ def __init__(self, header, peer):
...
+    async def __find_ancestor(self):
+        # n-ary search
+        start = max(self.root_state.tip.height - self.root_state.root_config.MAX_STALE_ROOT_BLOCK_HEIGHT_DIFF, 0)
+        end = min(self.root_state.tip.height, self.header.height)
+
+        while end >= start:
+            span = (end - start) // ROOT_BLOCK_HEADER_LIST_LIMIT + 1
+            resp = await self.__download_block_header_and_check(start, span - 1, len(range(start, end + 1, span)))
+
+            if len(resp.block_header_list) == 0:
```
qcdll: nit: `len(range(start, end + 1, span))` can probably just be `ROOT_BLOCK_HEADER_LIST_LIMIT`

quarkchain/cluster/master.py
```diff
@@ -107,119 +108,127 @@ def __init__(self, header, peer):
...
-                block_header_chain.append(header)
+            # Remote root block is reorg with new tip and new height (which may be lower than that of current)
+            # Setup end as the new height
+            if resp.root_tip != self.header:
+                self.header = resp.root_tip
 
-        block_header_chain.reverse()
+            if block_header_chain[0].hash_prev_block != ancestor.get_hash():
+                # TODO: Remote chain may reorg, may retry the sync
+                raise RuntimeError("Bad peer sending incorrect canonical headers")
```
qcdll: if this could be caused by legitimate reorg, maybe drop the "Bad peer" part

quarkchain/cluster/master.py
```diff
@@ -107,119 +108,127 @@ def __init__(self, header, peer):
...
-                block_header_chain.pop(0)
+                for block in block_chain:
+                    await self.__add_block(block)
+                    ancestor = block_header_chain[0]
+                    block_header_chain.pop(0)
 
     def __has_block_hash(self, block_hash):
         return self.root_state.db.contain_root_block_by_hash(block_hash)
 
-    def __validate_block_headers(self, block_header_list):
```
qcdll: cool!

quarkchain/cluster/master.py
qizhou: end  - start may be smaller than ROOT_BLOCK_HEADER_LIST_LIMIT (e.g., end == start) so we should not ask for maximum headers in each request.

qcdll: ah, right

quarkchain/cluster/master.py
qizhou: Done.

PR #519, opened 2019-04-20
add network id to discovery message

qcdll:
so that we won't pollute our own p2p discovery


quarkchain/p2p/discovery.py
```diff
@@ -158,7 +158,11 @@ def __init__(self,
         self.neighbours_callbacks = CallbackManager()
         self.topic_nodes_callbacks = CallbackManager()
         self.parity_pong_tokens = {} # : Dict[Hash32, Hash32]
+        self.network_id = network_id
         self.cancel_token = CancelToken('DiscoveryProtocol').chain(cancel_token)
+        self.QKC_ID_STRING = QKC_ID_STRING_TEMPLATE.format(self.network_id).encode()
```
ninjaahhh: nit: lowercase for member variables

qcdll: would like to keep the old name from const

PR #521, opened 2019-04-22
Update deployment utility scripts

ninjaahhh:


mainnet/Dockerfile
```diff
@@ -35,7 +35,7 @@ WORKDIR /code
 
 # docker build --build-arg CACHEBUST=$(date +%s) .
 ARG CACHEBUST=1
-RUN git clone --branch v2.3.3 --single-branch --depth 1 https://github.com/QuarkChain/pyquarkchain.git
```
ninjaahhh: `--single-branch --depth 1` will break the client version polling introduced in #512

quarkchain/tools/config_p2p.py
```diff
@@ -0,0 +1,38 @@
+"""
+    python config_p2p.py <64 character hex>
```
ninjaahhh: was thinking finding a way to combine `config_slave.py`, `config_p2p.py` and `miner_address.py`

will do in another day

PR #522, opened 2019-04-22
add get minor block headers with skip p2p command

qizhou:
Implement server-side command to get minor block headers from the canonical chain - either by height and hash.  May replace existing minor block synchronizer logic in future (not the plan for mainnet launch).

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -128,6 +128,44 @@ class PeerShardConnection(VirtualConnection):
             self.shard_state.root_tip, self.shard_state.header_tip, header_list
         )
 
+    async def handle_get_minor_block_header_list_with_skip_request(self, request):
+        if request.branch != self.shard_state.branch:
+            self.close_with_error("Wrong branch from peer")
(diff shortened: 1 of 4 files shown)


quarkchain/cluster/shard.py
```diff
@@ -128,6 +128,44 @@ def broadcast_tx_list(self, tx_list):
...
+            block_height = request.get_height()
+        else:
+            block_hash = request.get_hash()
+            block_header = self.shard_state.db.get_minor_block_header_by_hash(block_hash)
+            if block_header is None:
+                return GetMinorBlockHeaderListResponse(self.shard_state.root_tip, self.shard_state.header_tip, [])
+
+            # Check if it is canonical chain
+            block_height = block_header.height
+            if self.shard_state.db.get_minor_block_header_by_height(block_height) != block_header:
```
ninjaahhh: note that our minor chain may have higher probability of being on a fork

qizhou: True.  But it depends on the locality of the shard BPs.   This is more a server-side code for future optimization, and we may not replace the current sync code at the moment.

PR #525, opened 2019-04-22
don't serialize token id with 0 balance

qizhou:
A small fix to be consistent with TokenBalanceMap


ninjaahhh:
lgtm, but seems it broke lots of tests

PR #526, opened 2019-04-22
add root block synchronizer fast path

qizhou:
Don't call find_ancestor() if we could determine the ancestor immediately.

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -137,12 +137,16 @@ class SyncTask:
         return resp
 
     async def __find_ancestor(self):
+        # Fast path
+        if self.header.hash_prev_block == self.root_state.tip.get_hash():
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/master.py
```diff
@@ -198,13 +198,17 @@ def __init__(self, header, peer, stats, root_block_header_list_limit):
         if self.__has_block_hash(self.header.get_hash()):
             return
 
-        ancestor = await self.__find_ancestor()
-        if ancestor is None:
-            raise RuntimeError(
-                "Cannot find common ancestor with max fork length {}".format(
-                    self.max_staleness
+        # Fast path
+        if self.header.hash_prev_block == self.root_state.tip.get_hash():
```
ninjaahhh: how about putting this logic inside `__find_ancestor()`?

PR #533, opened 2019-04-24
Fix several issues of __find_ancestor()

qizhou:
- check() is too rigid
- find_ancestor should return best ancestor instead of None


qizhou:
> nit: `prevHeader` -> `prev_header`

where is prevHeader?

ninjaahhh:
> > nit: `prevHeader` -> `prev_header`
> 
> where is prevHeader?

line 168

qizhou:
> > > nit: `prevHeader` -> `prev_header`
> > 
> > 
> > where is prevHeader?
> 
> line 168

done

PR #541, opened 2019-04-29
Enforce order of inited shard IDs

ninjaahhh:
we actually rely on the order such that full shard ID is monotonically increasing during validation while python dict won't guarantee the order.

we are just being [lucky](https://stackoverflow.com/a/40007169) that the insertion order is preserved.


ninjaahhh:
> LGTM. Curious which part of code depends on the order assumption of ids?

1. when we are mining the root block, `MasterServer.__create_root_block_to_mine` will call this function and pack minor blocks into one root block. 
2. during `RootState.validate_block`, we assume the packed minor blocks are sorted by their full shard ID

qizhou:
Got it.  Nice catch!

PR #563, opened 2019-05-29
Support p2p-discovery-only mode

ninjaahhh:


quarkchain/cluster/cluster.py
```diff
@@ -89,7 +89,9 @@ def __init__(self, config, cluster_id=""):
 
     async def run(self):
         await self.run_master()
```
qizhou: Should we also run master in discovery mode?

ninjaahhh: yes we should, since the p2p network is started inside `master.py`

PR #566, opened 2019-06-01
Fix limit in tx history endpoint with token

ninjaahhh:


quarkchain/cluster/shard_db_operator.py
```diff
@@ -113,24 +113,26 @@ def get_transactions_by_address(
 
         tx_list = []
         for k, v in self.db.reversed_range_iter(start, end):
-            limit -= 1
             if limit < 0:
```
skji:
```suggestion
             if limit < 1:
```

skji: Tested in local cluster :)

ninjaahhh: good catch

Issue #552, opened 2019-05-05
Improve transaction pool


ninjaahhh:
two main features:

1. better sorting (gas price, and may also take multi native token into account)
2. support pending tx with higher nonce, instead of throwing them away

for you to triage @qizhou

ninjaahhh:
for dealing with tx with really low gas fee, should we 1) enforce a minimum gas price, or 2) simply enforce a limit to the size of tx queue or add a "self-cleaning" method to discard low-fee tx if reaching a threshold?

ninjaahhh:
for sorting by multi-native token gas price, a separate issue is created as  #577 which is a stretch goal

PR #571, opened 2019-06-11
Support future tx in tx queue

ninjaahhh:
#552


ninjaahhh:
tested on local devnet

Issue #575, opened 2019-06-17
`max_root_blocks_in_memory` too high


ninjaahhh:
currently 22500 * 2 blocks, which means on starting-up the memory usage would be really high (~8gb) and makes the cluster more prone to OOM

this property is affected by `MAX_STALE_ROOT_BLOCK_HEIGHT_DIFF` config

ninjaahhh:
cc @qizhou

ninjaahhh:
per the discussion, following approaches can alleviate the problem

1. decouple `MAX_STALE_ROOT_BLOCK_HEIGHT_DIFF` (which dictates to what degree we should discard an old block) and number of blocks to load into memory during recovery
1. add eviction policy for those in-memory hash -> block/header/meta pool

PR #581, opened 2019-06-18
Update recovery blk # and read rb through db

ninjaahhh:
addresses #575 

1. max stale block number is separated from number of blocks to recover
1. removed consistency check, `r_header_pool` is cache


ninjaahhh:
tested on a new node syncing from latest snapshot and seems working fine

PR #583, opened 2019-06-20
Remove root state header cache

ninjaahhh:
similar as #582, can rely on db-level cache (not configured yet)

 related to #575


ninjaahhh:
together with #582 and #581, a node will start with 1.5g memory and climb to 2.5g in one day (no mining), from which we can conclude with certain confidence:

1. `r_header_pool` and `m_header_pool` definitely used lots of memory
1. but eliminating them all won't save the world as memory still creeps up gradually

another good thing is I tried start a node syncing from scratch with latest code, syncing will take about 1 day but has no consistency problem; however the memory usage significantly increases (around 5g).

PR #584, opened 2019-06-21
Fix committing rb in failure

ninjaahhh:
during root block insertion, when root block failed to be broadcasted to slaves, local db still has its record which will prevent re-addition.


516108736:
nice

ninjaahhh:
invalid since `check` should guarantee the success of `ADD_ROOT_BLOCK` RPC, otherwise the cluster should crash since it's been trapped into a inconsistent state.

PR #589, opened 2019-06-27
add run_cluster.sh

qizhou:


ninjaahhh:
lol I would suggest keeping this simple script locally instead of in the repo

qizhou:
> lol I would suggest keeping this simple script locally instead of in the repo

I think the script is for all users rather than us or me.  Essentially, we should have a script that
- Detect whether the minor address is set or not.  If not, prompt user to set it.
- Detect whether the root tip is about 1 week earlier than the current time, if so prompt user to download the db
- Run the cluster

Basically for a new user, it just
- docker pull ....
- docker run ... (this may be also in a script, and call run_cluster.sh)

If will be great if a single script can perform both pull (latest version) and docker run (and run_cluster.sh)

PR #590, opened 2019-06-27
add query_balance.py

qizhou:
A simple code to query all balance of a shard chain at a specific block.


qizhou:
> nit: we may want to query the balance at a specific minor block height

This will be a good beginner task.

PR #594, opened 2019-06-28
Add min gas price threshold

### LINKED ISSUE #574: Allow configuring minimum gas price for tx

ninjaahhh:
`MIN_TX_POOL_GAS_PRICE` dictates min gas price for accepting a transaction through jsonrpc and p2p. similar to `--txpool.pricelimit` in goethereum

`MIN_MINE_GAS_PRICE`  dictates the min gas price for packing tx to a block for mining. similar to `--gasprice` in goethereum

closes #574


quarkchain/cluster/jsonrpc.py
```diff
@@ -684,7 +684,7 @@ def get_data_default(key, decoder, default=None):
         tx = TypedTransaction(SerializedEvmTransaction.from_evm_tx(evm_tx))
         success = await self.master.add_transaction(tx)
         if not success:
-            return None
+            return "0x" + bytes(32 + 4).hex()
```
qizhou: from core import Constant

bytes(Constant.ADDRESS_LENGTH)

quarkchain/tests/test_config.py
```diff
@@ -181,8 +182,12 @@ def test_serialization(self):
     "BLOCK_REWARD_DECAY_FACTOR": 0.5,
     "ENABLE_TX_TIMESTAMP": null,
     "TX_WHITELIST_SENDERS": [],
-    "ENABLE_EVM_TIMESTAMP": null
-}"""
+    "ENABLE_EVM_TIMESTAMP": null,
+    "TX_POOL_PRICE_THRESHOLD": 1000000000000000000,
```
qizhou: I should suggest using

MIN_TX_POOL_GAS_PRICE

or

MIN_TX_POOL_GAS_PRICE_THRESHOLD

as THRESHOLD can be THRESHOLD for lower or greater values

qizhou: Similar to the previous comment, would suggestion
MIN_MINE_TX_GAS_PRICE

or

MIN_MINE_TX_GAS_PRICE_THRESHOLD

PR #588, opened 2019-06-27
Separate config switch for enabling JSONRPC

hanyunxu:
Fix #570

PR #597, opened 2019-06-28
Fix jsonrpc switch

hanyunxu:
Follow up on #588


quarkchain/cluster/master.py
```diff
@@ -1535,8 +1535,8 @@ def main():
 
     # p2p discovery mode will disable master-slave communication and JSONRPC
     start_master = not env.cluster_config.P2P.DISCOVERY_ONLY
-    public_json_rpc_enabled = not env.cluster_config.ENABLE_PUBLIC_JSON_RPC
```
ninjaahhh: we can probably save those two variables and inline them into line 1559 and 1563 directly

PR #600, opened 2019-07-01
optimize root block processing time by caching LRUed root blocks

qizhou:


qizhou:
- improve db check time from 15s to 3.5s (about 100 root blocks)

quarkchain/cluster/root_state.py
```diff
@@ -404,13 +415,8 @@ def is_same_chain(self, longer_block_header, shorter_block_header):
 
     def validate_block(self, block):
         """Raise on validation errors """
-        if block.header.version != 0:
-            raise ValueError("incorrect root block version")
-
-        if not self.db.contain_root_block_by_hash(block.header.hash_prev_block):
```
ninjaahhh: why remove this check?

qizhou: Because all the check is now in validate_block_header()

PR #603, opened 2019-07-01
Discard underpriced tx in queue

### LINKED ISSUE #573: Discard underpriced tx when queue full
```python
    def add_tx(self, tx: TypedTransaction, xshard_gas_limit=None):
        if (
            len(self.tx_queue)
            > self.env.quark_chain_config.TRANSACTION_QUEUE_SIZE_LIMIT_PER_SHARD
        ):
            # exceeding tx queue size limit
            return False
    ......
```

for now we simply ignore new tx. a better way would be to throw away underpriced tx.

ninjaahhh:
closes #573 

note `bisect.insort` has amortized O(n) cost, ~9x slower than heapq push, may need optimization if becoming a bottleneck


qizhou:
I would suggest adding some performance comments here in case we found performance issue later.

PR #609, opened 2019-07-02
add 8 node mainnet config

qizhou:
Add 8 node config, slightly faster than 4 node config in 8-core machine
- 100 rblocks (and mblocks) processing time takes about 30s from 38s


ninjaahhh:
have you run it in pypy?

qizhou:
To my surprise, using pypy3 is even slower than python3 (takes 54.55s).

PR #613, opened 2019-07-03
simplify cross-shard processing

qizhou:
put x-shard processing code in top-level function apply_transaction rather than lower-level functions.


ninjaahhh:
need to fix the patch

PR #620, opened 2019-07-09
add db check rblock range

qizhou:
db check now accepts the rblock range to check so that if we find an issue in a rblock (together with mblocks), we could quickly reply it.

### DIFF
--- a/quarkchain/cluster/cluster.py
+++ b/quarkchain/cluster/cluster.py
@@ -31,15 +31,9 @@ def kill_child_processes(parent_pid, sig=signal.SIGTERM):
         process.send_signal(sig)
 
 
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/cluster.py
```diff
@@ -31,12 +31,15 @@ def kill_child_processes(parent_pid, sig=signal.SIGTERM):
         process.send_signal(sig)
 
 
-async def run_master(config_file, check_db, profile):
+async def run_master(config_file, check_db, profile, rblock_from, rblock_to):
```
ninjaahhh: prefer having `rblock_from` and `rblock_to` together as a string arg or dict arg

qizhou: Done.

PR #624, opened 2019-07-09
xshard withdraw checks posw balance

qizhou:
Previous code breaks db check with the following command:

QKC__QUARKCHAIN__DISABLE_POW_CHECK=True ./run_cluster.sh --check_db=true --check_db_rblock_from=105470 --check_db_rblock_to=105460


qizhou:
Now the same command passes db check lol

PR #625, opened 2019-07-10
increase rblock/mblock cache size and add mblock header cache

qizhou:
check db faces performance issue around rblock 15850, which takes several minutes per 10 blocks.  Adding the optimization will reduce the time to a few seconds.


quarkchain/cluster/shard_db_operator.py
```diff
@@ -344,7 +345,11 @@ def get_total_tx_count(self, m_block_hash):
         return int.from_bytes(count_bytes, "big")
 
     def get_minor_block_header_by_hash(self, h) -> Optional[MinorBlockHeader]:
+        if h in self.mblock_header_cache:
+            return self.mblock_header_cache[h]
         block = self.get_minor_block_by_hash(h)
+        if block is not None:
+            self.mblock_header_cache[h] = block.header
```
ninjaahhh:
```
if block is not None:
  self.mblock_header_cache[h] = block.header
  return block.header
return None

Issue #567, opened 2019-06-03
P2P routing table population takes a long time


ninjaahhh:
verified 6 bootnodes are functioning as expected (correctly bonded through discovery protocol ping/pong), however the actual routing table needs much more time to populate, varying between 1 to 10 minutes

needs some investigation

ninjaahhh:
restarting helps

ninjaahhh:
as in Kademlia spec, a node will contact boot node first and ask for neighbors then try to bond with them. in some of our trials, all the neighbors will timeout during bonding, thus the node will repeat the process indefinitely (ask boot node for neighbors -> fail bonding with neighbors -> ask boot node again).

we suspect the reason is the boot node failed to clean up failed nodes. will investigate more

ninjaahhh:
the reason is for bootnodes, the k-bucket is never full, so it rarely evicts bad nodes. we can make the node pruning more aggressive.

ninjaahhh:
deployed #632 to some of the bootnodes, will observe how the pruning behaves in production before merging the PR

ninjaahhh:
https://github.com/QuarkChain/pyquarkchain/pull/632#issuecomment-510969566

ninjaahhh:
close for now, until we see more problems

PR #632, opened 2019-07-11
Prune p2p routing table in bootnodes

ninjaahhh:
related to #567


ninjaahhh:
hmm after inspecting those deployed bootnodes, their routing tables are all stuck at size 50.

turned on debugging log on one of them, will observe further

PR #635, opened 2019-07-12
Add batch processing for check db

qizhou:
check db will submit the check tasks to a slave by collecting mheaders from multiple rblocks.  This reduces time that slaves maybe idle while master is still checking rblock.

Performance improvement (109295 rblocks)
Before:  4566.5641s
After:  3676.9211s

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -2,6 +2,7 @@ import argparse
 import asyncio
 import os
 import cProfile
+import time
 
 import psutil
 import time
@@ -902,6 +903,7 @@ class MasterServer:
             self.shutdown()
             raise Exception("Integrity check failure!")
 
+        start_time = time.monotonic()
         # Start with root db
         rb = self.root_state.get_tip_block()
         check_db_rblock_from = self.env.arguments.check_db_rblock_from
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/master.py
```diff
@@ -902,61 +903,85 @@ def log_error_and_exit(msg):
...
+                    )
 
-            try:
-                self.root_state.add_block(rb, write_db=False, skip_if_too_old=False)
-            except Exception as e:
-                Logger.log_exception()
-                log_error_and_exit(
-                    "Failed to check root block height {}".format(rb.header.height)
-                )
+            for crb in rb_list:
```
ninjaahhh: ideally this part (line 959 - 968) can be done in parallel as waiting for slave responses (line 970 - 978). I would assume the latter takes more time because of the amount of minor blocks to handle in a slave

quarkchain/cluster/master.py
```diff
@@ -902,61 +903,85 @@ def log_error_and_exit(msg):
...
+            rb_list = []
+            for i in range(self.env.arguments.check_db_rblock_batch):
+                count += 1
+                if rb.header.height < max(check_db_rblock_to, 1):
+                    break
+                rb_list.append(rb)
+                prev_rb = self.root_state.db.get_root_block_by_hash(
+                    rb.header.hash_prev_block
                 )
+                if prev_rb.header.height != rb.header.height - 1:
```
ninjaahhh: shouldn't this and line 938 be part of `root_state.validate_block`?

PR #636, opened 2019-07-12
Simplify get_transaction_by_address()

hanyunxu:
Chain id is unnecessary  for matching because no txs in other chains is stored.

### DIFF
--- a/quarkchain/cluster/shard_db_operator.py
+++ b/quarkchain/cluster/shard_db_operator.py
@@ -143,7 +143,7 @@ class TransactionHistoryMixin:
 
         # only recipient addr needed to match
         # chain id can be ignored since no TX on other chains is stored here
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/shard_db_operator.py
```diff
@@ -141,8 +141,9 @@ def get_transactions_by_address(
         if not self.env.cluster_config.ENABLE_TRANSACTION_HISTORY:
             return [], b""
 
-        serialized_address = address.serialize()
-        end = b"index_addr_" + serialized_address[:-2]
+        # only recipient addr needed to match
+        # chain id can be ignored since no TX on other chains is stored here
```
ninjaahhh: `chain id` -> `full shard key` can be ignored
`chain` -> `shard`

PR #640, opened 2019-07-15
adjust gas price to constantinople

qizhou:
- CALL gas from 40 to 700
- ExpBytes from 10 to 50
- Balance from 20 to 400
- SLOAD from 50 to 200
- SELFDESTRUCT from 0 to 5000

code to reference:

https://github.com/ethereum/go-ethereum/blob/master/params/gas_table.go


qizhou:
Actually, this doesn't change existing behavior - it just improves the readability of the existing code.

PR #643, opened 2019-07-16
Bring back EVM test + Various Fixes

ninjaahhh:
1. use mock serialization (aka old-school Ethereum RLP) for accounts
2. tweak EVM testing workflow


quarkchain/evm/state.py
```diff
@@ -519,14 +540,22 @@ def commit(self, allow_empties=False):
...
-                        acct.token_balances.serialize(),
-                        acct.storage,
-                        acct.code_hash,
-                        acct.full_shard_key,
-                        b"",
-                    )
+                    if self.is_testing:
+                        _acct = _MockAccount(
+                            acct.nonce,
+                            acct.token_balances.balance(token_id_encode("QKC")),
```
qizhou: I would suggest to add an assertion that other token balances are zero.

quarkchain/evm/state.py
```diff
@@ -248,6 +258,7 @@ def __init__(
         qkc_config=None,
         executing_on_head=False,
         db=None,
+        is_testing=False,
```
qizhou: Using the name is_testing may be confusing.  I would suggest naming it to "use_conv_evm_account" or something like that.

PR #653, opened 2019-07-22
unify mk_contract_address usage

qizhou:


quarkchain/evm/transactions.py
```diff
@@ -204,7 +199,7 @@ def intrinsic_gas_used(self):
     def creates(self):
         "returns the address of a contract created by this tx"
         if self.to in (b"", "\0" * 20):
-            return mk_contract_address(self.sender, self.nonce)
+            return mk_contract_address(self.sender, self.nonce, self.to_full_shard_key)
```
ninjaahhh: this property is not being used. I suggest we can simply remove it

qizhou: Done.

PR #660, opened 2019-07-26
Enforced EVM check on non-default token ID

czy1234:


fixtures/QuarkChainStateTests/CallPrecompile.json
```diff
@@ -0,0 +1,102 @@
+{
+    "CallPrecompile" : {
+        "_info" : {
+            "comment" : "",
+            "filledwith" : "testeth 1.6.0-alpha.0-11+commit.978e68d2",
```
ninjaahhh: those info are false

fixtures/QuarkChainStateTests/CallPrecompile.json
```diff
@@ -0,0 +1,102 @@
...
+                    "0x1234": "0x0de0b6b3a7640000"
+                },
+                "code" : "",
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            },
+            "0x0c2e77121daf0270d26bf0a7e9ab0faa8bf739ef" : {
+                "balance" : "0x00",
+                "code" : "608060405260043610603e5763ffffffff7c010000000000000000000000000000000000000000000000000000000060003504166336091dff81146043575b600080fd5b605f60048036036020811015605757600080fd5b503515156095565b6040518082602080838360005b838110156082578181015183820152602001606c565b5050505090500191505060405180910390f35b609b60c8565b811560be5760026040826000808064514b430001600019f1151560bc575060015b505b6001600055919050565b602060405190810160405280600190602082028038833950919291505056fea165627a7a72305820879435732afe1698d54a1957988b9e5a9f2da0a1eea0646c312bd3befbd5fd1a0029",
```
ninjaahhh: comments? with only the byte code I don't understand how it works

fixtures/QuarkChainStateTests/CallPrecompile.json
```diff
@@ -0,0 +1,102 @@
...
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            }
+        },
+        "transaction" : {
+            "data" : [
+                "0x36091dff0000000000000000000000000000000000000000000000000000000000000001",
```
ninjaahhh: explain what those data mean

quarkchain/evm/messages.py
```diff
@@ -557,14 +567,15 @@ def _apply_msg(ext, msg, code):
                 want=msg.value,
             )
             # TODO: Why return success if the transfer failed?
-            return 1, msg.gas, []
-
-    if msg.transfer_token_id != ext.default_state_token:
-        # TODODLL calling smart contract with non QKC transfer_token_id is not supported
-        return 1, msg.gas, []
+            return 0, msg.gas, []
```
ninjaahhh: note this change may break the consensus. need to make sure it conforms to our current history by using the db check tools.

on the other hand, seems this should never happen because transfer failure should be caught during validation. nevertheless, to be safe, we should run this change against db check tools

quarkchain/evm/messages.py
```diff
@@ -557,14 +567,15 @@ def _apply_msg(ext, msg, code):
...
 
     # Main loop
     if msg.code_address in ext.specials:
+        if (
+            msg.code_address
+            == b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x51\x4b\x43\x00\x01"
```
ninjaahhh: a constant variable should help readability

quarkchain/evm/tests/new_statetest_utils.py
```diff
@@ -106,7 +106,6 @@ def compute_state_test_unit(state, txdata, indices, konfig, is_qkc_state, qkc_en
             data=decode_hex(remove_0x_head(txdata["data"][indices["data"]])),
             gas_token_id=token_id_encode("QKC"),
             transfer_token_id=transfer_token_id,
-            # Should not set testing flag if testing QuarkChain state
```
ninjaahhh: why remove?

quarkchain/evm/state.py
```diff
@@ -278,6 +278,7 @@ def __init__(
         self.sender_disallow_map = dict()  # type: Dict[bytes, int]
         self.shard_config = ShardConfig(ChainConfig())
         self.use_mock_evm_account = use_mock_evm_account
+        self.cur_token_id_called = False
```
ninjaahhh: nit: naming it to `token_id_queried`?

quarkchain/evm/tests/new_statetest_utils.py
```diff
@@ -186,6 +184,8 @@ def init_state(env, pre, is_qkc_state, qkc_env=None):
                 big_endian_to_int(decode_hex(k[2:])),
                 big_endian_to_int(decode_hex(v[2:])),
             )
+    if seen_token_ids:
```
ninjaahhh: seems duplicate with line 190 to 191

quarkchain/evm/tests/new_statetest_utils.py
```diff
@@ -210,6 +210,7 @@ def verify_state_test(test):
             continue
         print("Testing for %s" % config_name)
         for result in results:
+            print(result)
```
ninjaahhh: remove

quarkchain/evm/messages.py
czy1234: Would you mind elaborating a bit on how to use the db check tools?

quarkchain/evm/tests/new_statetest_utils.py
czy1234: My mistake sorry about that

quarkchain/evm/messages.py
```diff
@@ -559,12 +575,10 @@ def _apply_msg(ext, msg, code):
...
             return 1, msg.gas, []
 
-    if msg.transfer_token_id != ext.default_state_token:
-        # TODODLL calling smart contract with non QKC transfer_token_id is not supported
-        return 1, msg.gas, []
-
     # Main loop
     if msg.code_address in ext.specials:
+        if msg.code_address == PROC_CURRENT_MNT_ID:
+            ext.token_id_queried = True
```
ninjaahhh: a second thought: you should probably do this inside `proc_current_mnt_id`

quarkchain/evm/specials.py
```diff
@@ -237,6 +244,7 @@ def proc_ecpairing(ext, msg):
         b"0000000000000000000000000000000000000006": proc_ecadd,
         b"0000000000000000000000000000000000000007": proc_ecmul,
         b"0000000000000000000000000000000000000008": proc_ecpairing,
+        b"000000000000000000000000000000514b430001": proc_current_mnt_id,
```
ninjaahhh: use the constant variable you just defined

quarkchain/evm/tests/new_statetest_utils.py
```diff
@@ -159,7 +162,6 @@ def init_state(env, pre, is_qkc_state, qkc_env=None):
         gas_limit=parse_int_or_hex(env["currentGasLimit"]),
         timestamp=parse_int_or_hex(env["currentTimestamp"]),
         qkc_config=qkc_env.quark_chain_config,
-        # If testing QuarkChain states, should not use mock account
```
ninjaahhh: should probably keep this comment?

fixtures/QuarkChainStateTests/CallPrecompile.json
```diff
@@ -0,0 +1,109 @@
...
+                    "0x8bb0": "0x0de0b6b3a7640000",
+                    "0x1234": "0x0de0b6b3a7640000"
+                },
+                "code" : "",
+                "nonce" : "0x00",
+                "storage" : {
+                }
+            },
+            "0x0c2e77121daf0270d26bf0a7e9ab0faa8bf739ef" : {
+                "_comment": "This is a contract with a function to call the precompiled contract PROC_CURRENT_MNT_ID",
```
ninjaahhh: I prefer to have the contract written somewhere, an example from existing test case in ethereum/tests looks like following

```
"comment" : "The first test case required here https://github.com/ethereum/tests/issues/431#issue-306081539 .\nInvoked Solidity 0.4.21 with solc --bin-runtime --optimize\ncontract DoubleSelfdestructTest {\n  address constant B1 = 0x1;\n  address constant B2 = 0x2;\n  function remoteSelfdestruct() external {\n    selfdestruct(B2);\n  }\n  function() external payable {\n    this.remoteSelfdestruct();\n    selfdestruct(B1);\n  }\n}\n",
```

fixtures/QuarkChainStateTests/CallPrecompile.json
```diff
@@ -1,7 +1,7 @@
 {
     "CallPrecompile" : {
         "_info" : {
-            "comment" : "",
+            "comment" : "pragma solidity ^0.5.0;\n\ncontract CallPrecompiled {\n    \n    uint256 x;\n    \n\tfunction test (bool flag) public payable returns(uint256[1] memory id){\n\t    if (flag) {\n\t \t    assembly {\n\t \t        let x := 2\n\t  \t        if iszero(call(not(0), 0x514b430001, 0, 0, 0x00, id, 0x40)) {\n\t      \t        x := 1\n\t   \t        }\n\t\t    }\n\t    }\n\t\tx = 1;\n\t}\n\t\n}",
```
ninjaahhh: the `x` inside the condition check is confusing. consider following

```
pragma solidity ^0.5.0;

contract CallPrecompiled {

    uint256 x = 0;

	function test (bool flag) public payable returns(uint256[1] memory id){
	    if (flag) {
	 	    assembly {
	 	        let _ := call(not(0), 0x514b430001, 0, 0, 0x00, id, 0x40)
		    }
	    }
		x = 1;
	}

}
```

also, why the state hashes are all different? should only have 2 results, right?

flag true && default token: x = 1
flag true && non-default token: x = 1
flag false && default token: x = 1
flag false && non-default token: x = 0 (failure case)

ninjaahhh:
more test cases (which deserves a separate file):

**`transfer` triggers fallback method, non-default transfer token ID invocation should succeed**

```solidity
pragma solidity >=0.5.0 <0.7.0;

contract Callee {
  function() external payable {
    uint256[1] memory id;
    assembly {
      let _ := call(not(0), 0x514b430001, 0, 0, 0x00, id, 0x40)
    }
  }
}

contract Caller {
  uint256 private state = 0;

  function call(Callee c) public payable {
    address(c).transfer(msg.value);
    state = 1;
  }
}
```

**similarly, `delegatecall` should also change the flag**

```solidity
pragma solidity >=0.5.0 <0.7.0;

contract Callee {
  uint256 public state = 0;
  function f() public {
    uint256[1] memory id;
    assembly {
      let _ := call(not(0), 0x514b430001, 0, 0, 0x00, id, 0x40)
    }
    state = 1;
  }
}

contract Caller {
  uint256 public state = 0;

  function call(Callee c) public payable {
    (bool success,) = address(c).delegatecall("f");
    require(success);
  }
}
```

quarkchain/evm/messages.py
```diff
@@ -578,6 +589,15 @@ def _apply_msg(ext, msg, code):
...
         ext.revert(snapshot)
+    if (
+        msg.depth == 0
+        and code
+        and msg.transfer_token_id != ext.default_state_token
+        and not ext.token_id_queried
+    ):
+        log_msg.debug("REVERTING")
+        ext.revert(snapshot)
+        return 0, 0, []
```
ninjaahhh: how about return unused gas instead of all the remaining gas?

quarkchain/evm/messages.py
```diff
@@ -217,7 +217,17 @@ def apply_transaction_message(
 
     gas_used = evm_gas_start - gas_remained + gas_used_start
 
-    if not result:
+    if (
+        message.transfer_token_id != ext.default_state_token
+    ) and not ext.token_id_queried:
```
ninjaahhh: should remove this part?

quarkchain/evm/messages.py
```diff
@@ -578,6 +583,15 @@ def _apply_msg(ext, msg, code):
     if res == 0:
         log_msg.debug("REVERTING")
         ext.revert(snapshot)
+    if (
```
qizhou: I think we should reuse the previous code as

if res == 0 and code != b"" and msg.transfer_token_id != ext.default_state_token and not ext.token_id_quieried:
  res = 1

qizhou: Also may skip checking if transfer value is zero.

quarkchain/evm/tests/new_statetest_utils.py
```diff
@@ -119,7 +119,10 @@ def compute_state_test_unit(state, txdata, indices, konfig, is_qkc_state, qkc_en
         # Run it
         prev = state.to_dict()
         success, output = apply_transaction(state, tx, tx_wrapper_hash=bytes(32))
-        print("Applied tx")
+        if success == 1:
+            print("Applied tx")
```
qizhou: Remove all print or replace them to log

quarkchain/evm/messages.py
```diff
@@ -575,6 +581,15 @@ def _apply_msg(ext, msg, code):
             post_storage=ext.log_storage(msg.to),
         )
 
+    if (
+        res == 1
+        and code != b""
+        and msg.transfer_token_id != ext.default_state_token
+        and not ext.token_id_queried
```
qizhou: I think we should check msg.token_id_queried.

ninjaahhh:
as discussed offline, please add some test cases on contract creation inside contracts

quarkchain/evm/specials.py
```diff
@@ -218,6 +218,7 @@ def proc_ecpairing(ext, msg):
 
 
 def proc_current_mnt_id(ext, msg):
+    msg.token_id_queried = True
```
qizhou: Do we still need this line?

PR #678, opened 2019-08-03
Change file path for edit coinbase

HangyuYe:
Call miner_address.py without export command may case wrong templet edited.


quarkchain/tools/miner_address.py
```diff
@@ -7,7 +7,7 @@
 from utils import underline, colorify
 from quarkchain.utils import sha3_256
 
-FILE = "../../testnet/2.5/cluster_config_template.json"
+FILE = "../mainnet/singularity/cluster_config_template.json"
```
ninjaahhh: wait, I think the path should be relative to current script, so should be `../../mainnet/xxx`?

HangyuYe: You are right! there are one more dir before pyquarkchain

Issue #652, opened 2019-07-22
tx history indexing out of order


ninjaahhh:
following is a screenshot of devnet

![image](https://user-images.githubusercontent.com/38544039/61666653-ee01b900-ac8c-11e9-8056-27aa6f6bcf12.png)

transactions should be ordered by block height (hence timestamp), but in the screenshot they are clearly not

the reason being when we range-query the index, we ignored full shard keys to "normalize" an address's full history, but the unexpected consequence is that full shard key will dominate the ordering of txs

a potential fix is to ignore full shard key when we save the index

PR #682, opened 2019-08-07
Fix tx history ordering

hanyunxu:
For #652


ninjaahhh:
confirmed by having a new node to re-index txs and the order looks normal

PR #685, opened 2019-08-07
Improve miner work cache to reduce stale block

### LINKED ISSUE #651: Vaild pow no reward
Hi.
I have been testing the cluster. So I used Nanominer and Ethminer(QKC) every time miner send vaild pow to the cluster, so cluster accepted it but no reward. The last test was ten valid pows; only 7 got the reward. Is it something wrong with the cluster? Even build docker image from GitHub and tested- the same results

ninjaahhh:
should fix #651

### DIFF
--- a/quarkchain/cluster/miner.py
+++ b/quarkchain/cluster/miner.py
@@ -257,10 +257,13 @@ class Miner:
 
         if now is None:  # clock open for mock
             now = time.time()
-        tip = self.get_header_tip_func()
+        work_prev_hash = (
+            self.current_work.header.hash_prev_block if self.current_work else None
+        )
+        tip_hash = self.get_header_tip_func().get_hash()
         if (
             not self.current_work  # no cache
-            or self.current_work.header.height <= tip.height  # cache outdated
+            or work_prev_hash != tip_hash  # cache outdated
             or now - self.current_work.header.create_time > 5  # stale
         ):
             block = await self.create_block_async_func(retry=False)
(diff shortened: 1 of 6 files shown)


quarkchain/cluster/miner.py
```diff
@@ -254,8 +257,12 @@ def _mine_new_block_async(self):
 
         if now is None:  # clock open for mock
             now = time.time()
-        # 5 sec interval magic number
-        if not self.current_work or now - self.current_work.header.create_time > 5:
+        tip = self.get_header_tip_func()
+        if (
+            not self.current_work  # no cache
+            or self.current_work.header.height <= tip.height  # cache outdated
```
qizhou: if tip is reset by root block, shard chain's tip may have smaller height.  I think we should check

or self.current_work.header.prev_has != tip.get_hash()

to make sure we are working on the correct tip + 1 block.

quarkchain/cluster/miner.py
```diff
@@ -294,6 +301,13 @@ def _mine_new_block_async(self):
         # this copy is necessary since there might be multiple submissions concurrently
         block = copy.deepcopy(self.work_map[header_hash])
         header = block.header
+
+        # reject if tip updated
+        tip = self.get_header_tip_func()
+        if header.height <= tip.height:
```
qizhou: Same comment as before.

quarkchain/cluster/miner.py
ninjaahhh: done

quarkchain/cluster/miner.py
ninjaahhh: done

ninjaahhh:
> Great! May we create a taste branch and ask our community to test?

we can cherry pick this commit and make a pre-release for ppl testing

PR #688, opened 2019-08-09
Fix xshard receipt log generation

ninjaahhh:
otherwise may have out of index errors when generating receipts with logs


quarkchain/cluster/filter.py
```diff
@@ -87,8 +87,15 @@ def _get_logs(self, blocks: List[MinorBlock]) -> List[Log]:
         """Given potential blocks, re-run tx to find exact matches."""
         ret = []
         for b_i, block in enumerate(blocks):
-            for i in range(len(block.tx_list or [])):
-                r = block.get_receipt(self.db.db, i)
+            deposit_hlist = (
```
qizhou: I would suggest adding one line comment to illustrate missing case (old version db)

quarkchain/cluster/shard_state.py
```diff
@@ -1548,7 +1548,13 @@ def get_transaction_receipt(
         if not block:
             return None
         try:
-            receipt = block.get_receipt(self.evm_state.db, index)
+            deposit_hlist = None
+            if index >= len(block.tx_list):
+                deposit_hlist = self.db.get_xshard_deposit_hash_list(
```
qizhou: deposit_hlist may be None

quarkchain/cluster/filter.py
ninjaahhh: done. put the older version db handling into the `block. get_receipt`, by returning a fake tx receipt

quarkchain/cluster/shard_state.py
ninjaahhh: done. similar to previous note

PR #689, opened 2019-08-11
add PoSWRental.sol

qizhou:
A simple contract to offer posw rental.

### DIFF
--- a/contracts/PoSWRental.sol
+++ b/contracts/PoSWRental.sol
@@ -23,12 +23,12 @@ contract PoSWRental{
         principal_ = principal_ - amount;
     }
 
-    function GetDivident() public view returns (uint256) {
+    function GetDividend() public view returns (uint256) {
         return address(this).balance - principal_;
     }
 
     function Payout() public {
-        uint256 amount = GetDivident() * minerRate_ / 10000;
+        uint256 amount = GetDividend() * minerRate_ / 10000;
 
         // Ignore failure if miner is a contract
         miner_.send(amount);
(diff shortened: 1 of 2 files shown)


contracts/PoSWRental.sol
```diff
@@ -0,0 +1,44 @@
...
+
+    function GetDivident() public view returns (uint256) {
+        return address(this).balance - principal_;
+    }
+
+    function Payout() public {
+        uint256 amount = GetDivident() * minerRate_ / 10000;
+
+        // Ignore failure if miner is a contract
+        miner_.send(amount);
```
ninjaahhh: comment says miner should not be a contract, but we are not enforcing miner being non-contract (by checking its `code`)

qizhou:
We could probably come up with a kind of interface design like ERC20 for PoSW rental and provide a reference code.

contracts/PoSWRental.sol
qizhou: I don't we could enforce as the miner address could be contract address but the contract is not created yet (then bypass the codesize == 0 check).  I think there is a similar security issue we discussed before.  As a result, it is the miner's responsibility to ensure the address is a user one.

ninjaahhh:
proposing a new impl

https://gist.github.com/ninjaahhh/4e7feaaba384d15b4a11c3c6c83cbb15

API

- fallback payable for adding stakes
- `withdrawStakes(amount)` for stakers to withdraw (including profit)
- `withdrawFee` for miners to get mining fees
- `updateMiner` for miners to update address

ninjaahhh:
close as a duplicate of QuarkChain/quarkchain-contracts#9

PR #692, opened 2019-08-14
A posw rental pool contract

qizhou:
Reference only, not audited.

The same interface as PoSWRental.sol with the following interface:

function AddPrincipal() payable external
function WithdrawPrincipal(uint amount) public 
function Payout() public 
function SetMiner(address payable miner) public

We may also add the following function
function WithdrawAll() public


qizhou:
Other features we have may have
- minimum principal to join
- staker may withdraw at payout (prevent principal to be locked forever).

PR #693, opened 2019-08-14
Fix xshard cursor neighbor check

ninjaahhh:
root cause: a shard may not have another shard's minor block's previous root block hash

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -174,7 +174,8 @@ class XshardTxCursor:
                 mblock_header.hash_prev_root_block
             )
             if (
-                prev_root_header.height
+                prev_root_header is None
+                or prev_root_header.height
                 <= self.shard_state.env.quark_chain_config.get_genesis_root_height(
                     self.shard_state.full_shard_id
                 )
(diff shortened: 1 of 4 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -174,7 +174,8 @@ def get_next_tx(self):
                 mblock_header.hash_prev_root_block
             )
             if (
-                prev_root_header.height
+                prev_root_header is None
```
qizhou: Looks like this is caused by the race condition, but I am still not sure where the race comes?

ninjaahhh:
> Looks like there is change related to the issue, but also changes in peer.py which seems to be not related. Could we separate them?

sure, will do

quarkchain/cluster/shard_state.py
ninjaahhh: suppose shard 1's root genesis height is 100

1. root block 100 contains a minor block from shard 0, and 
2. that minor block links to an older root block, say root block with height 50
3. then, when shard 1 receives the root block 100 and wants to generate its own minor block with height 1, it will go through root block 100's minor blocks, find out one of them links to root block 50 which doesn't exist in its db, and throw the exception

PR #701, opened 2019-08-17
Fix root block reward as tx apply

### LINKED ISSUE #666: Root chain coinbase reward should be not burnt by PoSW

ninjaahhh:
fix #666. unit tests will follow


ninjaahhh:
check-db passed. confirmed this won't break current consensus

PR #706, opened 2019-08-22
Support getWork with specified coinbase address

ninjaahhh:


quarkchain/cluster/miner.py
```diff
@@ -251,42 +261,39 @@ def _mine_new_block_async(self):
...
-        # clean up worker map
-        # TODO: for now, same param as go-ethereum. or LRU cache?
-        self.work_map = {
-            h: b
-            for h, b in self.work_map.items()
-            if now - b.header.create_time < 7 * 12
-        }
+            header_hash = block.header.get_hash_for_mining()
+            # note that key could be None
+            self.current_works[coinbase_addr] = header_hash
```
qizhou: If coinbase_addr is None, why not use configed coinbase?  Using None as the key is very confusing.

ninjaahhh: because otherwise we have to pass the default coinbase (the one from the config) to `class Miner`, which I think is unnecessary

PR #710, opened 2019-08-26
make qkchash to work without &&

qizhou:


qizhou:
Looks like we don't need it anymore.

PR #712, opened 2019-08-28
[Pub/Sub] Part3 - Subscription manager

czy1234:
- Tests need to be fixed

### DIFF
--- a/quarkchain/cluster/slave.py
+++ b/quarkchain/cluster/slave.py
@@ -878,6 +878,7 @@ class SlaveServer:
         # block hash -> future (that will return when the block is fully propagated in the cluster)
         # the block that has been added locally but not have been fully propagated will have an entry here
         self.add_block_futures = dict()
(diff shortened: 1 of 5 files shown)


quarkchain/cluster/slave.py
```diff
@@ -902,6 +903,9 @@ def remove_cluster_peer_id(self, cluster_peer_id):
             await shard.create_peer_shard_connections(
                 self.cluster_peer_ids, self.master
             )
+            self.shard_subscription_managers[
```
ninjaahhh: I don't like initializing subscription manager inside shard state, prefer moving this part to line 922

```
sub_manager = SubscriptionManager()
self.shard_subscription_managers[full_shard_id] = sub_manager
shard = Shard(self.env, full_shard_id, self, sub_manager)
new_shards.append(shard)
```

ninjaahhh: a second thought, current form is fine

PR #716, opened 2019-08-29
[Pub/Sub] Support newPendingTX and Syncing with subscription manager

hanyunx:
- Subscription manager will be rebased to #712 later.
- More unit tests to be added


ninjaahhh:
since #712 has been merged to master, can safely change comparing base to master now

ninjaahhh:
please resolve conflicts

quarkchain/cluster/shard.py
```diff
@@ -435,14 +435,18 @@ def __validate_block_headers(self, block_header_list: List[MinorBlockHeader]):
 class Synchronizer:
     """ Buffer the headers received from peer and sync one by one """
 
-    def __init__(self):
+    def __init__(self, instance):
```
ninjaahhh: this will work but I'm not a fan of those circular references between instances. how about only pass in `shard.state.subscription_manager.notify` function?

quarkchain/cluster/shard_state.py
```diff
@@ -1281,6 +1287,14 @@ def add_cross_shard_tx_list_by_minor_block_hash(
         - it is a neighor of current shard following our routing rule
         """
         self.db.put_minor_block_xshard_tx_list(h, tx_list)
+        for tx in tx_list.tx_list:
+            asyncio.ensure_future(
+                self.subscription_manager.notify(
+                    "newPendingTransactions",
```
ninjaahhh: prefer using a constant as in `quarkchain/constants.py`. the similar applies to the `SubscriptionManager`

quarkchain/cluster/tests/test_jsonrpc.py
```diff
@@ -1230,195 +1210,3 @@ def test_newPendingTransactions(self):
             self.assertEqual(results[0]["id"], 6)
             self.assertEqual(results[1]["params"]["subscription"], results[0]["result"])
             self.assertTrue(results[1]["params"]["result"], tx.get_hash())
-
-    def test_logs(self):
```
ninjaahhh: why remove if later it's still needed?

quarkchain/cluster/shard_state.py
ninjaahhh: also `add_cross_shard_tx_list_by_minor_block_hash` seems to be invoked in 2 different cases

1. broadcast xshard tx from current shard to other shards
2. handle broadcasted xshard tx from other shards

and seems we only need to care about the second case, because the first case is already handled by `add_tx` for notifying pending tx

hanyunx: Actually the pending tx response will not display twice as mentioned above empirically. For some reason (trying to figure out why), the subscription manager processed here is always target shard's, so only subscription of target shard will get notified.

quarkchain/cluster/subscription.py
```diff
@@ -57,14 +57,18 @@ def remove_subscriber(self, sub_id):
...
-        data = None
+        data = [running, tip, queue]
         await self.__notify(SUB_SYNC, data)
 
     async def __notify(self, sub_type, data):
         assert sub_type in self.subscribers
         for sub_id, websocket in self.subscribers[sub_type].items():
-            response = self.response_encoder(sub_id, data)
+            if sub_type == SUB_SYNC:
+                response == self.syncing_response_encoder(
```
czy1234: Maybe the data encoding process can be put in notify_sync? I guess this can make __notify() look cleaner.

hanyunx: We need to call `syncing_response_encoder ` specifically for syncing because of its different response format from other events.

quarkchain/cluster/shard_state.py
```diff
@@ -1281,6 +1286,13 @@ def add_cross_shard_tx_list_by_minor_block_hash(
         - it is a neighor of current shard following our routing rule
         """
         self.db.put_minor_block_xshard_tx_list(h, tx_list)
+        for tx in tx_list.tx_list:
```
ninjaahhh: as mentioned before, this function will also be called by the source shard, which we shouldn't do pub

quarkchain/cluster/subscription.py
```diff
@@ -74,3 +78,24 @@ def response_encoder(sub_id, result):
             "method": "subscription",
             "params": {"subscription": sub_id, "result": result},
         }
+
+    @staticmethod
+    def syncing_response_encoder(sub_id, running, tip, queue):
+        if not running:
```
ninjaahhh:
```
ret = {
                "jsonrpc": "2.0",
                "subscription": sub_id,
                "result": {"syncing": running},
}
if running:
  ret["result"]["status"] = ...
```

quarkchain/cluster/subscription.py
```diff
@@ -57,14 +57,18 @@ def remove_subscriber(self, sub_id):
     async def notify_sync(
         self, running: bool, tip: MinorBlockHeader, queue: List[MinorBlockHeader]
     ):
-        # TODO
-        data = None
+        data = [running, tip, queue]
         await self.__notify(SUB_SYNC, data)
 
     async def __notify(self, sub_type, data):
         assert sub_type in self.subscribers
```
ninjaahhh: always move if-check outside the loop if possible
``` 
encoder = self.response_encoder if sub_type == SUB_SYNC else self.sync_status_encoder
for ...
  response = encoder(sub_id, data)
```

quarkchain/cluster/subscription.py
```diff
@@ -74,3 +78,24 @@ def response_encoder(sub_id, result):
             "method": "subscription",
             "params": {"subscription": sub_id, "result": result},
         }
+
+    @staticmethod
+    def syncing_response_encoder(sub_id, running, tip, queue):
```
ninjaahhh: nit: how about naming as `sync_status_encoder`?

quarkchain/cluster/shard_state.py
ninjaahhh: oh just saw your reply. please investigate why only the target shards run this code path

ninjaahhh:
also make sure it handles chain reorg correctly. from [infura doc](https://infura.io/docs/ethereum/wss/eth_subscribe)

> Returns the hash for all transactions that are added to the pending state and are signed with a key that is available in the node. When a transaction that was previously part of the canonical chain isn't part of the new canonical chain after a reogranization its again emitted.

quarkchain/cluster/shard_state.py
hanyunx: Will do.

quarkchain/cluster/shard.py
```diff
@@ -435,17 +435,23 @@ def __validate_block_headers(self, block_header_list: List[MinorBlockHeader]):
 class Synchronizer:
     """ Buffer the headers received from peer and sync one by one """
 
-    def __init__(self, notify_sync, tip):
+    def __init__(self, notify_sync, header_tip_getter: Callable[[], MinorBlockHeader]):
```
ninjaahhh: also add type hinting for `notify_sync`

quarkchain/cluster/subscription.py
```diff
@@ -55,20 +55,21 @@ def remove_subscriber(self, sub_id):
...
 
     async def notify_sync(
-        self, running: bool, tip: MinorBlockHeader, queue: List[MinorBlockHeader]
+        self,
+        running: bool,
+        header_tip: MinorBlockHeader,
+        highest_block: MinorBlockHeader,
     ):
-        data = [running, tip, queue]
+        data = [running, header_tip, highest_block]
```
ninjaahhh: you don't need pass the whole header, only the height is enough

quarkchain/cluster/subscription.py
```diff
@@ -80,22 +81,13 @@ def response_encoder(sub_id, result):
...
-                "status": {
-                    "startingBlock": tip,
-                    "highestBlock": max(h.height for h, _ in queue),
-                },
-            },
-        }
+        ret = {"jsonrpc": "2.0", "subscription": sub_id, "result": {"syncing": running}}
+        if running:
+            ret["result"]["status"] = {
+                "startingBlock": header_tip,
```
ninjaahhh: from https://ethereum.stackexchange.com/questions/31145/what-does-startingblock-mean-in-eth-syncing

height of header tip should be current block, and I guess starting block should be the minimum height among the queue?

hanyunx: ah make sense!

mainnet/singularity/cluster_config_template.json
```diff
@@ -466,6 +466,7 @@
     {
       "HOST": "127.0.0.1",
       "PORT": 38001,
+      "WEBSOCKET_JSON_RPC_PORT": 38591,
```
ninjaahhh: don't update the mainnet config

quarkchain/cluster/subscription.py
```diff
@@ -55,16 +55,17 @@ def remove_subscriber(self, sub_id):
         await self.__notify(SUB_LOGS, data)
 
     async def notify_sync(
-        self, running: bool, tip: MinorBlockHeader, queue: List[MinorBlockHeader]
-    ):
-        # TODO
-        data = None
+        self, data: Optional[Tuple[int, ...]] = None
+    ):  # data = (header_tip_height, highest_block)
```
ninjaahhh: remove comments

quarkchain/cluster/tests/test_jsonrpc.py
```diff
@@ -1181,89 +1181,342 @@ def send_websocket_request(request, num_response=1, port=38590):
...
+            self.assertTrue(slaves[0].add_tx(tx))
+
+            tx_responses = [json.loads(call_async(websocket.recv())) for _ in requests]
+            for i, resp in enumerate(tx_responses):
+                self.assertEqual(
+                    resp["params"]["subscription"], sub_responses[i]["result"]
+                )
+                self.assertTrue(resp["params"]["result"], tx.get_hash())
+
+    def test_new_pending_tx_with_reorg(self):
```
ninjaahhh: use `add_tx` to follow the common path

PR #724, opened 2019-09-04
Update EVM log filter

czy1234:


quarkchain/cluster/log_filter.py
```diff
@@ -58,8 +58,8 @@ def __init__(
     def _get_block_candidates(self) -> List[MinorBlock]:
         """Use given criteria to generate potential blocks matching the bloom."""
         ret = []
-        for i in range(self.start_block, self.end_block + 1):
-            block = self.db.get_minor_block_by_height(i)
+        for i in range(self.size):
+            block = self.db.get_minor_block_by_hash(self.end_block_hash)
```
ninjaahhh: use `get_minor_block_header_by_hash` since we only need the header

quarkchain/cluster/log_filter.py
```diff
@@ -81,6 +81,8 @@ def _get_block_candidates(self) -> List[MinorBlock]:
             if (1 + i) % 100 == 0 and time.time() - self.start_ts > Filter.TIMEOUT:
                 raise Exception("Filter timeout")
 
+            self.end_block_hash = block.header.hash_prev_block
```
ninjaahhh: don't do this. use a new variable

quarkchain/cluster/shard_state.py
```diff
@@ -6,7 +6,7 @@
 
 from rlp import DecodingError
 
-from quarkchain.cluster.filter import Filter
+from quarkchain.cluster.log_filter import Filter
```
ninjaahhh: also rename `Filter` to `LogFilter`?

quarkchain/cluster/log_filter.py
```diff
@@ -32,8 +32,8 @@ def __init__(
         self.db = db
         # if `addresses` present, should be in the same shard
         self.recipients = [addr.recipient for addr in addresses]
-        self.start_block = start_block
-        self.end_block = end_block
+        self.end_block_hash = end_block_hash
```
ninjaahhh: should probably pass the block header directly

PR #725, opened 2019-09-05
Add precompile contract to deploy root chain staking contract

ninjaahhh:
needed to have a deterministic staking contract address

for more details, check QuarkChain/QCEPs#2


quarkchain/evm/specials.py
```diff
@@ -254,6 +256,27 @@ def proc_transfer_mnt(ext, msg):
...
+        0,
+        msg.gas - gascost,
+        ROOT_CHAIN_POSW_CONTRACT_BYTECODE,
+        msg.depth + 1,
+        to_full_shard_key=msg.to_full_shard_key,
+        transfer_token_id=msg.transfer_token_id,
+        gas_token_id=msg.gas_token_id,
+    )
+    # CREATE2, deterministic staking contract address
+    return create_contract(ext, new_msg, b"", salt=bytes(32))
```
qizhou: Why not create a contract with a specific address like precompiled contracts?

quarkchain/constants.py
```diff
@@ -26,3 +26,9 @@
...
+""".strip()
+)
```
qizhou: I will suggest adding a field ROOT_CHAIN_POSW_CONTRACT_BYTECODE_HASH and use the field to check in shard state

quarkchain/evm/specials.py
ninjaahhh: then the handling logic of the tx targeting at precompiled address will be kinda complicated, something like this:

```
# messages.py

        res, gas, dat = ext.specials[msg.code_address](ext, msg)
    else:
        res, gas, dat = vm.vm_execute(ext, msg, code)

# specials.py

def proc(ext, msg):
  if ext.has_code(msg.to):
    return vm.vm_execute(ext, msg, code)
  else:
    # make a deployment message
```

which complicates the logic a little bit and I don't see much benefit here, except that the address is slightly easier to remember

quarkchain/constants.py
ninjaahhh: will do

ninjaahhh: update: I added the bytecode hash to config, but leaves the actual bytecode in the constants

PR #739, opened 2019-09-16
Add args for setting websocket port

hanyunx:


quarkchain/cluster/cluster_config.py
```diff
@@ -265,6 +265,12 @@ def attach_arguments(cls, parser):
 
         parser.add_argument("--num_slaves", default=4, type=int)
         parser.add_argument("--port_start", default=38000, type=int)
+        parser.add_argument(
+            "--websocket_port_start",
+            default=38590,
+            type=int,
+            help="if empty, will start from 38590 by default",
```
ninjaahhh: wait, I think the correct default behavior is to not start websocket on slave servers

PR #742, opened 2019-09-17
Fix evm state for getting root chain stakes

ninjaahhh:


quarkchain/cluster/shard_state.py
```diff
@@ -1888,12 +1888,16 @@ def get_minor_block_by_height(
         return block, posw_info and posw_info._asdict()
 
     def get_root_chain_stakes(
-        self, recipient: bytes, block_hash: bytes
+        self,
+        recipient: bytes,
+        block_hash: bytes,
+        mock_evm_state: Optional[EvmState] = None,  # open for testing
```
ninjaahhh: cc @syntrust

ninjaahhh:
tested on testnet3, syncing works

quarkchain/cluster/shard_state.py
```diff
@@ -1888,12 +1888,16 @@ def get_minor_block_by_height(
...
     ) -> (int, bytes):
-        h = self.db.get_minor_block_header_by_hash(block_hash)
-        check(h is not None)
-        evm_state = self._get_evm_state_from_height(h.height).ephemeral_clone()
-        evm_state.gas_used = 0
+        meta = self.db.get_minor_block_meta_by_hash(block_hash)
+        check(meta is not None)
+        evm_state = mock_evm_state or self.__create_evm_state(
```
qizhou: If mock_evm_state is not Null, should we still check meta is not None?

ninjaahhh: good point. I think we can skip checking meta when mocking, but as for now the tests are not broken, seems no need to change the code

PR #747, opened 2019-09-18
Fix transfer failure return result

ninjaahhh:
in normal case this code path won't be touched because:

1. normal transfer tx will be guarded by validation
2. transfer / calls in smart contract will be guarded by VM checking QKC balance

thus there are two ways to break the execution rule:

1. indirectly make a CALL msg (outside the VM) such as calling transfer_mnt precompiled contract (fixed by this PR)
~2. make a CALL message with value transfer but in non-default token~

### DIFF
--- a/fixtures/QuarkChainStateTests/stContractCallMnt/contractTransferMntNotEnoughBalance.json
+++ b/fixtures/QuarkChainStateTests/stContractCallMnt/contractTransferMntNotEnoughBalance.json
@@ -1,0 +1,77 @@
+{
+  "contractTransferMntNotEnoughBalance": {
+    "_info": {
+      "comment": "deployed contract: https://gist.github.com/hanyunx/b758b44b4c866a4c175acf5017761b13",
+      "filledwith": "byhand",
+      "lllcversion": "byhand",
+      "source": "NA",
+      "sourceHash": "NA"
+    },
+    "env": {
+      "currentCoinbase": "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
+      "currentDifficulty": "0x020000",
+      "currentGasLimit": "0xf5f5e100",
+      "currentNumber": "0x01",
+      "currentTimestamp": "0x03e8",
+      "previousHash": "0x5e20a0453cecd065ea59c37ac63e079ee08998b6045136a8ce6635c7912ec0b6"
+    },
+    "post": {
+      "ConstantinopleFix": [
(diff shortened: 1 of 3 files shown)


fixtures/QuarkChainStateTests/stContractCallMnt/contractTransferMntNotEnoughBalance.json
```diff
@@ -0,0 +1,77 @@
...
+      "ConstantinopleFix": [
+        {
+          "hash": "0xf5811632e06923cfbbd8b8d2fc8dcd13dd0c583f96d385f203d423fd73bee80c",
+          "indexes": {
+            "data": 0,
+            "gas": 0,
+            "value": 0,
+            "transferTokenId": 0
+          },
+          "logs": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"
```
ninjaahhh: no need for log

fixtures/QuarkChainStateTests/stContractCallMnt/contractTransferMntNotEnoughBalance.json
```diff
@@ -0,0 +1,77 @@
...
+        },
+        {
+          "hash": "0xf5811632e06923cfbbd8b8d2fc8dcd13dd0c583f96d385f203d423fd73bee80c",
```
ninjaahhh: same hash as first one. can remove?

ninjaahhh:
cc @pingke this is a consensus change for current testnet3, please make sure the new test can pass on your side

quarkchain/evm/messages.py
```diff
@@ -560,8 +560,7 @@ def _apply_msg(ext, msg, code):
                 have=ext.get_balance(msg.sender, token_id=msg.transfer_token_id),
                 want=msg.value,
             )
-            # TODO: Why return success if the transfer failed?
-            return 1, msg.gas, []
+            return 0, 0, []
```
qizhou: May we double check how goetherum works on this to make sure the behavior is consistent?  In addition, I am wondering if there is an existing test in eth repo to test it?

ninjaahhh: as mentioned in my first comment, in normal cases this should be dead code. however in our system it becomes possible in transfer_mnt precompiled contract (or if we have a bug on `apply_xshard_deposit`)

PR #752, opened 2019-09-20
Fix check db on root block pow

ninjaahhh:


quarkchain/cluster/master.py
```diff
@@ -994,8 +994,24 @@ def log_error_and_exit(msg):
 
             for crb in rb_list:
                 try:
+                    # lower the difficulty for root block signed by guardian
```
qizhou: Could we put a common code for master so that we have a single code path in master to work on root chain posw?

ninjaahhh: sounds good. addressed in #753

PR #756, opened 2019-09-23
Add enable ts for precompiled contracts

ninjaahhh:


quarkchain/evm/messages.py
```diff
@@ -564,7 +564,11 @@ def _apply_msg(ext, msg, code):
 
     # Main loop
     if msg.code_address in ext.specials:
```
qizhou: Why not use

special_proc, enable_ts = ext.specials.get(msg.code_address, (None, 0))
if special_proc is not None and ext.block_timestamp >= enable_ts:
...

ninjaahhh: good point. done

quarkchain/env.py
```diff
@@ -1,14 +1,16 @@
...
 from quarkchain.cluster.cluster_config import ClusterConfig
+from quarkchain.evm.specials import specials
 
 
 class Env:
-    def __init__(self, db=None, evm_config=None, cluster_config=None):
+    def __init__(self, db=None, evm_config=None):
         self.db = db or InMemoryDb()
-        self.cluster_config = cluster_config if cluster_config else ClusterConfig()
+        self.__cluster_config = ClusterConfig()
```
qizhou: Why not still use cluster_config and then apply timestamp in the ctor?  This makes sure we have the best backward-compatible interface.

ninjaahhh: some places we are manually setting the cluster config (like in `slave.py`):

```python
    env = DEFAULT_ENV.copy()
    env.cluster_config = ClusterConfig.create_from_args(args)
    env.slave_config = env.cluster_config.get_slave_config(args.node_id)
    env.arguments = args
```

qizhou: Then I would suggest using

self.cluster_config = ClusterConfig()

PR #766, opened 2019-10-04
Add timestamp when creating EVM state

hanyunx:
Before this PR, a transaction would fail if a user attempts to send a transaction at the moment of starting a cluster and not any peers has been connected. This is because the timestamp of EVM state is set as 0 by default, in which case tx is disabled except for whitelist senders.

As a potential solution, the timestamp is updated when creating the EVM state such that a transaction can be applied successfully, though it will not be broadcasted to other nodes and will eventually be discarded.

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -317,7 +317,7 @@ class ShardState:
         self.confirmed_header_tip = confirmed_header_tip
         sender_disallow_map = self._get_sender_disallow_map(header_tip)
         self.evm_state = self.__create_evm_state(
-            self.meta_tip.hash_evm_state_root, sender_disallow_map, time.time()
+            self.meta_tip.hash_evm_state_root, sender_disallow_map, int(time.time())
         )
         check(
             self.db.get_minor_block_evm_root_hash_by_hash(header_tip_hash)
@@ -1478,7 +1478,7 @@ class ShardState:
             b = self.db.get_minor_block_by_hash(h)
             sender_disallow_map = self._get_sender_disallow_map(b.header)
             evm_state = self.__create_evm_state(
-                b.meta.hash_evm_state_root, sender_disallow_map, time.time()
+                b.meta.hash_evm_state_root, sender_disallow_map, int(time.time())
             )
             self.__update_tip(b, evm_state)
             Logger.info(
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -565,7 +565,6 @@ class ShardState:
         )
         if ephemeral:
             state = state.ephemeral_clone()
-        state.timestamp = block.header.create_time
         state.gas_limit = block.header.evm_gas_limit
         state.block_number = block.header.height
         state.recent_uncles[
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -553,7 +560,9 @@ def _get_evm_state_for_new_block(self, block, ephemeral=True):
             prev_minor_header, recipient=coinbase_recipient
         )
 
-        state = self.__create_evm_state(root_hash, sender_disallow_map)
+        state = self.__create_evm_state(
+            root_hash, sender_disallow_map, block.header.create_time
```
ninjaahhh: can remove line 568

ninjaahhh:
as discussed before, please update the context for this task

hanyunx:
> as discussed before, please update the context for this task
done.

quarkchain/cluster/shard_state.py
```diff
@@ -317,7 +317,9 @@ def init_from_root_block(self, root_block):
         self.confirmed_header_tip = confirmed_header_tip
         sender_disallow_map = self._get_sender_disallow_map(header_tip)
         self.evm_state = self.__create_evm_state(
-            self.meta_tip.hash_evm_state_root, sender_disallow_map
+            self.meta_tip.hash_evm_state_root,
+            sender_disallow_map,
+            header_tip.create_time + 1,
```
ninjaahhh: maybe `time.time()` should be better?

quarkchain/cluster/shard_state.py
```diff
@@ -1472,7 +1480,9 @@ def add_root_block(self, root_block: RootBlock):
             b = self.db.get_minor_block_by_hash(h)
             sender_disallow_map = self._get_sender_disallow_map(b.header)
             evm_state = self.__create_evm_state(
-                b.meta.hash_evm_state_root, sender_disallow_map
+                b.meta.hash_evm_state_root,
+                sender_disallow_map,
+                b.header.create_time + 1,
```
ninjaahhh: same here

quarkchain/evm/state.py
```diff
@@ -703,6 +704,8 @@ def ephemeral_clone(self):
         s.journal = copy.copy(self.journal)
         s.cache = {}
         s.qkc_config = self.qkc_config
+        if timestamp:
```
ninjaahhh: how about simply do 

```
s.timestamp = self.timestamp
```

hanyunx: will remove the update of timestamp here since the timestamp has already been updated when creating the evm state while this is a copy of that state.

quarkchain/evm/state.py
```diff
@@ -1,5 +1,6 @@
 # Modified based on pyethereum under MIT license
 import copy
+from typing import Optional
```
ninjaahhh: not needed?

quarkchain/cluster/shard_state.py
```diff
@@ -317,7 +317,7 @@ def init_from_root_block(self, root_block):
         self.confirmed_header_tip = confirmed_header_tip
         sender_disallow_map = self._get_sender_disallow_map(header_tip)
         self.evm_state = self.__create_evm_state(
-            self.meta_tip.hash_evm_state_root, sender_disallow_map
+            self.meta_tip.hash_evm_state_root, sender_disallow_map, time.time()
```
ninjaahhh: realized `time.time()` returns float but the timestamp is usually int

ninjaahhh: please change to `int(time.time())`

PR #780, opened 2019-10-28
Filter out minor block headers with larger create time

hanyunx:
Solve https://github.com/QuarkChain/quarkchain-dev/issues/383

### DIFF
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -1271,10 +1271,6 @@ class MasterServer:
 
         adjusted_diff = await self.__adjust_diff(r_block)
 
-        for m_block_header in r_block.minor_block_header_list:
-            if m_block_header.create_time > r_block.header.create_time:
-                r_block.minor_block_header_list.remove(m_block_header)
-
         try:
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/master.py
```diff
@@ -1270,6 +1270,11 @@ def handle_new_root_block_header(self, header, peer):
         self.root_state.write_committing_hash(r_block.header.get_hash())
 
         adjusted_diff = await self.__adjust_diff(r_block)
+
+        for m_block_header in r_block.minor_block_header_list:
```
ninjaahhh: changing root block body may break the validation of `hash_merkle_root`

quarkchain/cluster/master.py
```diff
@@ -1270,6 +1270,7 @@ def handle_new_root_block_header(self, header, peer):
         self.root_state.write_committing_hash(r_block.header.get_hash())
 
         adjusted_diff = await self.__adjust_diff(r_block)
+
```
hanyunx: will remove in the next commit (if any).

ninjaahhh: remove it!

PR #781, opened 2019-11-05
Add precompiled contract for minting MNT

hanyunx:


quarkchain/evm/specials.py
```diff
@@ -281,6 +284,25 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
...
 
 
+# 3 inputs: (minter, token ID, amount)
+def proc_mint_mnt(ext, msg):
+    gascost = 3
+    if msg.gas < gascost:
+        return 0, 0, []
+
+    target_addr, _ = _system_contracts[SystemContract.MINT_MULTI_NATIVE_TOKEN]
+    assert (
```
ninjaahhh: ...you don't want to crash the program if the caller doesn't match

ninjaahhh:
also please add unit tests, which shouldn't be blocked by the system contract's bytecode

quarkchain/evm/vm.py
```diff
@@ -923,3 +923,4 @@ def __init__(self):
         self.create = lambda msg: 0, 0, 0
         self.call = lambda msg: 0, 0, 0
         self.sendmsg = lambda msg: 0, 0, 0
+        self._state = state
```
ninjaahhh: add a comment: this field is only for mocking purpose

quarkchain/evm/specials.py
```diff
@@ -322,7 +348,11 @@ def addr(self) -> bytes:
     SystemContract.ROOT_CHAIN_POSW: (
         decode_hex(b"514b430000000000000000000000000000000001"),
         ROOT_CHAIN_POSW_CONTRACT_BYTECODE,
-    )
+    ),
+    SystemContract.MINT_MULTI_NATIVE_TOKEN: (
+        decode_hex(b"514b430000000000000000000000000000000002"),
+        MINT_MULTI_NATIVE_TOKEN_CONTRACT_BYTECODE,
```
ninjaahhh: consider the naming to be in sync with `NonReservedNativeTokenManager`?

quarkchain/evm/specials.py
```diff
@@ -281,6 +284,24 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
     return create_contract(ext, new_msg, target_addr)
 
 
+# 3 inputs: (minter, token ID, amount)
+def proc_mint_mnt(ext, msg):
+    gascost = 3
+    if msg.gas < gascost:
+        return 0, 0, []
+
+    target_addr, _ = _system_contracts[SystemContract.MINT_MULTI_NATIVE_TOKEN]
```
ninjaahhh: `target_addr` looks misleading, how about `allowed_sender`?

quarkchain/evm/tests/test_specials.py
```diff
@@ -20,3 +22,20 @@ def test_current_mnt_id(self):
         self.assertListEqual([result, gas_remained], [1, 4 - 3])
         self.assertEqual(len(data), 32)
         self.assertEqual(int.from_bytes(data, byteorder="big"), 2 ** 64 - 1)
+
+    def test_proc_mint_mnt(self):
```
ninjaahhh: another test case would be some random user trying to mint native tokens but failed

quarkchain/evm/specials.py
```diff
@@ -281,6 +284,24 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
...
+    # Only authorized account has access to minting new token
+    if target_addr != msg.sender:
+        return 0, 0, []
+
+    minter = msg.data.extract32(0)
+    mnt = msg.data.extract32(32)
```
ninjaahhh: `mnt` to `token_id`?

quarkchain/evm/specials.py
```diff
@@ -281,6 +284,24 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
     return create_contract(ext, new_msg, target_addr)
 
 
+# 3 inputs: (minter, token ID, amount)
+def proc_mint_mnt(ext, msg):
+    gascost = 3
```
ninjaahhh: how much gas should we charge for updating token balance? @qizhou

quarkchain/evm/tests/test_specials.py
```diff
@@ -20,3 +23,36 @@ def test_current_mnt_id(self):
...
+        result, gas_remained, ret = proc_mint_mnt(VmExtBase(state), msg)
+        self.assertListEqual([result, gas_remained], [1, 5 - 3])
+        self.assertEqual(len(ret), 32)
+        self.assertEqual(int.from_bytes(ret, byteorder="big"), 1)
+
+        balance = state.get_balance(minter, int.from_bytes(token_id, byteorder="big"))
+        self.assertEqual(balance, int.from_bytes(amount, byteorder="big"))
+
+    def test_proc_mint_mnt_invalid_user(self):
+        addr = Address.create_random_account().recipient
```
ninjaahhh: the code is roughly the same as `test_proc_mint_mnt`, can you put them together? 

```python
testcases = [(sys_contract_addr, True), (random_addr, False)]
for addr, expect_success in testcases:
  ...
  if expect_success:
    ...
  else:
    ...
```

quarkchain/evm/specials.py
qizhou: I think it should be the same as CALL, where the gas is calculated as following

# Creating a new account
            if op == "CALL" and not ext.account_exists(to) and (value > 0):
                extra_gas += opcodes.GCALLNEWACCOUNT
            # Value transfer
            if value > 0:
                extra_gas += opcodes.GCALLVALUETRANSFER

since the case that "to" should already exist, we probably only need GCALLVALUETRANSFER,  which is 9000.

BTW: Looks like proc_transfer_mnt()'s gas price is underprice based on the same logic here.

quarkchain/evm/specials.py
```diff
@@ -281,6 +284,24 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
     return create_contract(ext, new_msg, target_addr)
 
 
+# 3 inputs: (minter, token ID, amount)
+def proc_mint_mnt(ext, msg):
+    gascost = opcodes.GCALLVALUETRANSFER
```
ninjaahhh: we should update the gas usage as follows

```
            # Extra gas costs based on various factors
            extra_gas = 0
            # Creating a new account
            if op == "CALL" and not ext.account_exists(to) and (value > 0):
                extra_gas += opcodes.GCALLNEWACCOUNT
            # Value transfer
            if value > 0:
                extra_gas += opcodes.GCALLVALUETRANSFER
```

quarkchain/evm/specials.py
```diff
@@ -286,7 +286,16 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
 
 # 3 inputs: (minter, token ID, amount)
 def proc_mint_mnt(ext, msg):
+    minter = msg.data.extract32(0)
+    token_id = msg.data.extract32(32)
+    amount = msg.data.extract32(64)
+    if amount == 0:
+        return 0, 0, []
+
     gascost = opcodes.GCALLVALUETRANSFER
```
ninjaahhh: as discussed offline, should follow the same condition check as "CALL"

ninjaahhh: ah nvm it's already been updated

quarkchain/evm/specials.py
```diff
@@ -293,7 +293,7 @@ def proc_mint_mnt(ext, msg):
         return 0, 0, []
 
     gascost = opcodes.GCALLVALUETRANSFER
-    if not ext.account_exists(minter):
+    if not ext._state.account_exists(minter):
```
ninjaahhh: `ext` has `account_exists` method

ninjaahhh: add a test case to cover different gas usage?

quarkchain/evm/specials.py
```diff
@@ -281,6 +284,30 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
...
+    if msg.gas < gascost:
+        return 0, 0, []
+
+    allowed_sender, _ = _system_contracts[SystemContract.NON_RESERVED_NATIVE_TOKEN]
+    # Only system contract has access to minting new token
+    if allowed_sender != msg.sender:
+        return 0, 0, []
+
+    ext._state.delta_token_balance(minter, token_id, amount)
+    return 1, msg.gas - gascost, [0] * 31 + [1]
```
syntrust: @ninjaahhh @hanyunx why return [0] * 31 + [1] here?

hanyunx: to indicate that token is successfully minted.

PR #784, opened 2019-11-11
gas utility evm

jyouyj:


quarkchain/core.py
```diff
@@ -1115,6 +1115,7 @@ class CrossShardTransactionDeposit(Serializable):
         ("gas_token_id", uint64),
         ("transfer_token_id", uint64),
         ("gas_remained", uint256),
+        ("refund_rate", uint256),
```
ninjaahhh: I think we don't need it if we can just convert the gas price

quarkchain/evm/messages.py
```diff
@@ -277,7 +286,7 @@ def apply_transaction_message(
     return success, output
 
 
-def apply_xshard_deposit(state, deposit, gas_used_start):
+def apply_xshard_deposit(state, deposit, gas_used_start, refund_rate):
```
ninjaahhh: same, can get rid of it

quarkchain/core.py
ninjaahhh: a second thought: converting gas price doesn't seem enough because we need the refund rate to determine how much should be sent to the miner and how much should be burned

cc @qizhou

quarkchain/evm/messages.py
```diff
@@ -351,7 +361,31 @@ def apply_transaction(state, tx: transactions.Transaction, tx_wrapper_hash):
         state.get_balance(tx.sender, token_id=tx.gas_token_id)
         >= tx.startgas * tx.gasprice
     )
-    state.delta_token_balance(tx.sender, tx.gas_token_id, -tx.startgas * tx.gasprice)
+    original_gas_token_id = tx.gas_token_id
+    refund_rate = 100
+    if tx.gas_token_id != state.genesis_token_id:
```
ninjaahhh: I don't think the code will work here, because only `ShardState` has `genesis_token_id`. can you add some unit tests?

quarkchain/evm/messages.py
```diff
@@ -351,7 +361,31 @@ def apply_transaction(state, tx: transactions.Transaction, tx_wrapper_hash):
...
+        refund_rate, gasprice = state.get_gas_utility_info(
```
ninjaahhh: `gasprice` can be easily confused with `tx.gasprice`. how about naming it as `converted_genesis_token_gas_price`?

PR #785, opened 2019-11-12
Support using trie for storing native token balances

### LINKED ISSUE #782: Support representing native tokens in trie
as we launch the auction for non-reserved native tokens, we may encounter a greater number of native tokens, which means the current storage scheme as a serialized dict is not enough

ninjaahhh:
resolves #782 

more tests to come

### DIFF
--- a/quarkchain/evm/state.py
+++ b/quarkchain/evm/state.py
@@ -111,7 +111,7 @@ class TokenBalances:
                 self.token_root = data[1:]
                 self.token_trie = SecureTrie(Trie(db, data[1:]))
             else:
-                raise Exception("Unknown enum byte in token_balances")
+                raise ValueError("Unknown enum byte in token_balances")
 
     def commit(self):
         self._committed = True
@@ -129,9 +129,9 @@ class TokenBalances:
                 self.token_trie.update(k, rlp.encode(bal))
             else:
                 self.token_trie.delete(k)
-                keys_to_del.append(k)
-        for k in keys_to_del:
-            del self._balances[k]
+                keys_to_del.append(token_id)
+        for token_id in keys_to_del:
+            del self._balances[token_id]
 
         self.token_root = self.token_trie.root_hash
 
@@ -146,13 +146,11 @@ class TokenBalances:
         # Serialize in-memory balance representation as an array
         if len(self._balances) == 0:
             return b""
-        ret = b"\x00"
         # Don't serialize 0 balance
         ls = [TokenBalancePair(k, v) for k, v in self._balances.items() if v > 0]
         # Sort by token id to make token balances serialization deterministic
         ls.sort(key=lambda b: b.token_id)
-        ret += rlp.encode(ls)
-        return ret
+        return b"\x00" + rlp.encode(ls)
 
     def balance(self, token_id):
         if token_id in self._balances:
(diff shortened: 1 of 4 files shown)


quarkchain/evm/state.py
```diff
@@ -604,15 +667,9 @@ def reset_storage(self, address):
...
 
     # Creates a snapshot from a state
-    def to_snapshot(self, root_only=False, no_prevblocks=False):
-        snapshot = {}
-        if root_only:
-            # Smaller snapshot format that only includes the state root
-            # (requires original DB to re-initialize)
-            snapshot["state_root"] = "0x" + encode_hex(self.trie.root_hash)
-        else:
-            # "Full" snapshot
```
ninjaahhh: "Full" snapshot is never being used in our code

quarkchain/evm/state.py
```diff
@@ -637,27 +694,7 @@ def to_snapshot(self, root_only=False, no_prevblocks=False):
     @classmethod
     def from_snapshot(cls, snapshot_data, env, executing_on_head=False):
         state = State(env=env)
-        if "alloc" in snapshot_data:
```
ninjaahhh: again, never used

ninjaahhh:
after introducing token trie, determining whether an account is empty becomes more difficult so I updated several call sites of checking whether the account is empty

quarkchain/evm/state.py
```diff
@@ -92,49 +92,120 @@ class TokenBalancePair(rlp.Serializable):
...
+    def set_balance(self, journal, token_id, value):
+        preval = self.balance(token_id)
+        self._balances[token_id] = value
+        journal.append(lambda: self._balances.__setitem__(token_id, preval))
+        if self._committed:
+            self._committed = False
+            journal.append(lambda: setattr(self, "_committed", True))
+
+    def is_blank(self):
+        return not self.token_trie and self._balances == {}
```
qizhou: We probably need _non_zero_entries == 0 to make sure zero balance, a further comment may be needed to explain if token_trie is not NULL_TRIE_HASH.

quarkchain/evm/state.py
```diff
@@ -92,49 +92,120 @@ class TokenBalancePair(rlp.Serializable):
...
+                    self._balances[p.token_id] = p.balance
+            elif enum == b"\x01":
+                self.token_root = data[1:]
+                self.token_trie = SecureTrie(Trie(db, data[1:]))
             else:
-                raise Exception("Unknown enum byte in token_balances")
+                raise ValueError("Unknown enum byte in token_balances")
+
+    def commit(self):
+        self._committed = True
```
qizhou: Similar to storage trie, I think we could remove _commit field here and clear _balances since _balances may never be used after commit

quarkchain/evm/state.py
```diff
@@ -92,49 +92,120 @@ class TokenBalancePair(rlp.Serializable):
...
+                self.token_trie.delete(k)
+                keys_to_del.append(token_id)
+        for token_id in keys_to_del:
+            del self._balances[token_id]
+
+        self.token_root = self.token_trie.root_hash
 
     def serialize(self):
-        if len(self.balances) == 0:
+        if not self._committed:
```
qizhou: Following the previous comment, I think we could check the commit status by checking
- token_trie is None or
- token_trie is not None and _balances = {}

quarkchain/evm/state.py
ninjaahhh: done

quarkchain/evm/state.py
ninjaahhh: updated. should check

1. token trie is None **and has less than 16 entries in balance cache**
2. token trie is not None and balance cache is {}

quarkchain/evm/state.py
ninjaahhh: makes sense. updated.

PR #787, opened 2019-11-14
Add deploy_system_contract function

hanyunx:

PR #788, opened 2019-11-14
deploy gas utility system contract

jyouyj:


quarkchain/evm/specials.py
```diff
@@ -281,6 +284,31 @@ def proc_deploy_root_chain_staking_contract(ext, msg):
     return create_contract(ext, new_msg, target_addr)
 
 
+def proc_deploy_multi_native_gas_utility_contract(ext, msg):
```
ninjaahhh: should use #787 once it's finished

PR #792, opened 2019-11-18
interact with gas utility system contract

jyouyj:


quarkchain/constants.py
```diff
@@ -39,6 +39,12 @@
 """.strip()
 )
 
+GENERAL_NATIVE_TOKEN_CONTRACT_BYTECODE = bytes.fromhex(
```
ninjaahhh: TODO: this is outdated, will update

PR #794, opened 2019-11-19
pay transaction gas with native token

jyouyj:
todo:

1. ~more tests~ 4c3f641279b1baf660e172df0a6b9e75b8905c62
1. ~update xshard deposit version handling~ #798
1. ~add validation in case gas reserve doesn't have enough balance~ #796


quarkchain/evm/messages.py
```diff
@@ -350,6 +363,28 @@ def apply_transaction(state, tx: transactions.Transaction, tx_wrapper_hash):
         state.get_balance(tx.sender, token_id=tx.gas_token_id)
         >= tx.startgas * tx.gasprice
     )
+    gasprice, refund_rate = tx.gasprice, 100
+    # convert gas if using non-genesis native token
+    if tx.gas_token_id != state.genesis_token:
+        refund_rate, converted_genesis_token_gas_price = pay_native_token_as_gas(
+            state, tx.gas_token_id, tx.startgas, tx.gasprice
+        )
+        assert converted_genesis_token_gas_price > 0
```
ninjaahhh: if contract call to `payAsGas` failed, should we mark this transaction failed? @qizhou

quarkchain/evm/messages.py
```diff
@@ -187,6 +188,16 @@ def validate_transaction(state, tx):
...
 
 
+def _refund(state, message, total, refund_percent):
+    to_refund = total * refund_percent // 100
+    to_burn = total - to_refund
+    state.delta_token_balance(message.sender, message.gas_token_id, to_refund)
+
+    # burn
+    if to_burn:
+        state.delta_token_balance(bytes(20), message.gas_token_id, to_burn)
```
qizhou: I would suggest using another constant address rather than bytes(20) to avoid confusion with special contract addresses.

PR #796, opened 2019-11-21
Update allowed native token and tx validation handling

### LINKED ISSUE #797: Failure handling when using native token to pay for gas
**background**:

when running a tx, if we find the gas token is different with default token (QKC), we will call the system contract's `payAsGas` function to update the contract storage and get back corresponding converted gas price in QKC, then deduct the system contract's QKC balance.

**problem**:

as asked in this PR https://github.com/QuarkChain/pyquarkchain/pull/794#discussion_r349363463, if the contract call failed or the system contract doesn't have enough QKC, how should we handle the transaction? note that:

1. it's not enough to pre-run the contract call during validation when adding the tx (i.e. before it's added to the block) to get the execution result and reject if failed: because the actual environment may be different when running the block v.s. validating the transaction
2. one goal of using native token to pay gas is to unify miners' reward tokens as QKC. so I think it's undesirable to simply mark the transaction failed and give the used gas to miner as a non-QKC token

**potential solutions**:

1. discard the block if such situation happens, i.e. force the node to make sure state is valid (e.g. system contract has enough QKC, contract call has to succeed) when adding transaction. however this needs us to update how we keep track of state when accepting transactions
1. fail the transaction, but no rewards to miner, burn part of the gas as dictated by `refund_rate` (however the miner will be less incentivized to accept these kind of transactions)

cc @qizhou

ninjaahhh:
resolves #797 

1. removed allowed token checks
2. do not allow non-default token transaction if balance is zero
3. add validation steps when using non-default token for paying gas: the gas reserve should be ready and have enough default tokens to cover the gas

### DIFF
--- a/quarkchain/cluster/tests/test_native_token.py
+++ b/quarkchain/cluster/tests/test_native_token.py
@@ -152,18 +152,21 @@ class TestNativeTokenShardState(unittest.TestCase):
         )
         self.assertFalse(state.add_tx(tx1))
 
-    @mock_pay_native_token_as_gas
+    @mock_pay_native_token_as_gas()
     def test_native_token_gas(self):
         """in-shard transfer QETH using native token as gas
         """
-        QETH = token_id_encode("QETH")
+        qeth = token_id_encode("QETH")
         id1 = Identity.create_random_identity()
         acc1 = Address.create_from_identity(id1, full_shard_key=0)
         acc2 = Address.create_random_account(full_shard_key=0)
+        # Miner
         acc3 = Address.create_random_account(full_shard_key=0)
 
         env = get_test_env(
-            genesis_account=acc1, genesis_minor_token_balances={"QETH": 10000000}
+            genesis_account=acc1,
+            genesis_minor_token_balances={"QETH": 10000000},
+            charge_gas_reserve=True,
         )
         state = create_default_shard_state(env=env)
 
@@ -174,8 +177,8 @@ class TestNativeTokenShardState(unittest.TestCase):
             to_address=acc2,
             value=12345,
             gas=21000,
-            gas_token_id=QETH,
-            transfer_token_id=QETH,
+            gas_token_id=qeth,
+            transfer_token_id=qeth,
         )
 
         self.assertTrue(state.add_tx(tx))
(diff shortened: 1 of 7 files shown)


quarkchain/cluster/tests/test_native_token.py
```diff
@@ -114,26 +114,7 @@ def test_native_token_transfer_0_value_success(self):
...
-        # MALICIOUS0 shall not be in the dict
-        self.assertNotEqual(
-            state.get_balances(acc1.recipient),
-            {self.genesis_token: 10000000 - opcodes.GTXCOST, MALICIOUS0: 0},
-        )
-        self.assertEqual(
-            state.get_balances(acc1.recipient),
-            {self.genesis_token: 10000000 - opcodes.GTXCOST},
-        )
+        self.assertFalse(state.add_tx(tx))
```
ninjaahhh: do not allow zero value transfer using non-default token

ninjaahhh:
@qizhou @hanyunx take a look at this PR when you have time

quarkchain/evm/messages.py
```diff
@@ -129,57 +131,60 @@ def validate_transaction(state, tx):
...
-        raise InsufficientBalance(
-            "{}: token {} is not in allowed transfer_token list".format(
-                tx.__repr__(), token_id_decode(tx.transfer_token_id)
+    default_chain_token = state.shard_config.default_chain_token
+    bal = {tx.transfer_token_id: state.get_balance(tx.sender, tx.transfer_token_id)}
+    if tx.transfer_token_id != tx.gas_token_id:
+        bal[tx.gas_token_id] = state.get_balance(tx.sender, tx.gas_token_id)
+
+    # (4) requires non-zero balance for transfer_token_id and gas_token_id if non-default
+    for token_id in [tx.transfer_token_id, tx.gas_token_id]:
```
qizhou: what is the purpose of this?  Looks like this should be a protocol of the upper layer.

ninjaahhh: as discussed offline, this will prevent senders using arbitrary token IDs to confuse other users (e.g. a tx with non-existent token id KLQJDLASD)

PR #798, opened 2019-11-23
Update cross-shard deposit schema

ninjaahhh:
tweaked serialization/deserialization of xshard tx deposit because of schema change

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -1193,8 +1193,9 @@ class CrossShardTransactionList(Serializable):
         self.version = 1  # force the version
 
     @staticmethod
-    def get_version(data, size) -> int:
+    def get_version(data) -> int:
         """-1 for deprecated list without version."""
+        size = ByteBuffer(data).get_uint(4)
         if (
             size == 0
             or (len(data) - 4) == CROSS_SHARD_TRANSACTION_DEPOSIT_DEPRECATED_SIZE * size
@@ -1204,8 +1205,7 @@ class CrossShardTransactionList(Serializable):
 
     @classmethod
     def from_data(cls, data):
-        size = ByteBuffer(data).get_uint(4)
-        version = cls.get_version(data, size)
+        version = cls.get_version(data)
 
         if version == 1:
             return cls.deserialize(data)
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -1102,7 +1102,7 @@ class CrossShardTransactionDepositDeprecated(Serializable):
         self.is_from_root_chain = is_from_root_chain
 
 
-class CrossShardTransactionDeposit(Serializable):
+class CrossShardTransactionDepositV0(Serializable):
     """ Destination of x-shard tx
     """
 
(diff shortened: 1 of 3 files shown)


quarkchain/evm/messages.py
```diff
@@ -407,102 +406,95 @@ def apply_transaction(state, tx: transactions.Transaction, tx_wrapper_hash):
...
-                state.qkc_config.ENABLE_EVM_TIMESTAMP is None
-                or state.timestamp >= state.qkc_config.ENABLE_EVM_TIMESTAMP
-            ):
-                remote_gas_reserved = tx.startgas - intrinsic_gas
-            ext.add_cross_shard_transaction_deposit(
-                quarkchain.core.CrossShardTransactionDeposit(
-                    tx_hash=tx_wrapper_hash,
-                    from_address=quarkchain.core.Address(
-                        tx.sender, tx.from_full_shard_key
+    if not tx.is_cross_shard:
```
ninjaahhh: changed the order of control flow for better readability. no actual change to the logic

quarkchain/core.py
```diff
@@ -1102,7 +1102,7 @@ def __init__(
         self.is_from_root_chain = is_from_root_chain
 
 
-class CrossShardTransactionDeposit(Serializable):
+class CrossShardTransactionDepositV0(Serializable):
```
qizhou: For versioning, I would suggest
- CrossShardTransactionDepositDeprecated as V0
- CrossShardTransactionDeposit as V1
- CrossShardTransactionDeposit with the refund field as V2

quarkchain/core.py
```diff
@@ -1162,49 +1171,76 @@ def __init__(self, tx_list):
...
-                    create_contract=False,
-                    is_from_root_chain=tx.is_from_root_chain,
-                    refund_rate=100,
-                )
-                for tx in old_list.tx_list
-            ]
-        )
+    @classmethod
+    def from_data(cls, data):
+        size = ByteBuffer(data).get_uint(4)
```
qizhou: This will throw if data is empty?

quarkchain/core.py
```diff
@@ -1162,49 +1171,76 @@ def __init__(self, tx_list):
...
+                        is_from_root_chain=tx.is_from_root_chain,
+                        refund_rate=100,
+                    )
+                    for tx in old_list.tx_list
+                ]
+            )
+
+        if version == 0:
+            v0_list = CrossShardTransactionListV0.deserialize(data)
+            # magic!
```
qizhou: I think we should use the similar logic for DeprecatedList for consistency.

quarkchain/core.py
ninjaahhh: it's the same logic as previous `from_data`: `is_old_list` will also read `size` first. so I think it probably will not throw

quarkchain/core.py
ninjaahhh: since we only added one field for V2, I think this is more readable and concise

quarkchain/core.py
ninjaahhh: good call. will do

ninjaahhh: wait, it doesn't make sense, as in line 1182 we are currently writing version 0 to all the deposits

quarkchain/core.py
ninjaahhh: updated

ninjaahhh:
updated, take another look when you have time @qizhou

PR #801, opened 2019-12-06
Fix mock paying native token as gas

ninjaahhh:
the meta decorator needs to be invoked to have the desired effect. thanks to it, I've identified other failed test cases


ninjaahhh:
cc @516108736

PR #804, opened 2019-12-12
Update xshard deposit from native token tests

ninjaahhh:
covered xshard tx using native token test cases

### DIFF
--- a/quarkchain/evm/specials.py
+++ b/quarkchain/evm/specials.py
@@ -271,10 +271,15 @@ def proc_deploy_system_contract(ext, msg):
 
     data = msg.data.extract32(0)
     contract_index = data if data else 1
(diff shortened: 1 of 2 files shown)


quarkchain/evm/specials.py
```diff
@@ -271,10 +271,15 @@ def proc_deploy_system_contract(ext, msg):
 
     data = msg.data.extract32(0)
     contract_index = data if data else 1
-    if contract_index not in [e.value for e in SystemContract]:
+
+    try:
+        target_addr, bytecode, enable_ts = _system_contracts[
+            SystemContract(contract_index)
+        ]
+    except (ValueError, KeyError) as e:
```
hanyunx: looks like `as e` is unnecessary

ninjaahhh: done

PR #805, opened 2019-12-12
Setup for Testnet3.2 

hanyunx:
- Add `ENABLE_GENERAL_NATIVE_TOKEN_TIMESTAMP` 
- Add config file for testnet
- Update bytecode with supervisor hardcoded


quarkchain/config.py
```diff
@@ -311,6 +311,7 @@ class QuarkChainConfig(BaseConfig):
     ENABLE_EVM_TIMESTAMP = None
     ENABLE_QKCHASHX_HEIGHT = None
     ENABLE_MNT_AUCTION_TIMESTAMP = None
+    ENABLE_GENERAL_NATIVE_TOKEN_TIMESTAMP = None
```
ninjaahhh: "enable non-reserved token manager timestamp" and "enable general token manager timestamp"?

hanyunx: `ENABLE_GENERAL_NATIVE_TOKEN_TIMESTAMP` is now consistent with system contract's name, while the former one can be changed to `ENABLE_NON_RESERVED_NATIVE_TOKEN_TIMESTAMP` to be consistent.

PR #808, opened 2019-12-26
add reorg_detector.py

qizhou:
Add a reorg detector to check a recently-created root block is reorganized or not.

### DIFF
--- a/quarkchain/tools/reorg_detector.py
+++ b/quarkchain/tools/reorg_detector.py
@@ -1,0 +1,111 @@
+import jsonrpcclient
+import time
+import logging
+import argparse
+import smtplib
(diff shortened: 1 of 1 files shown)


ninjaahhh:
this can only detect **current** reorg (happening right now, when the script is running), but not history re-orgs, right?

qizhou:
Depending on the definition of history, a user could setup rblock_height_diff and check_interval with different values to adjust sensitivity:
- rblock_height_diff = 0, and check_interval = 30: This basically detects stale block rate together with possible reorg
- rblock_height_diff = 10, and check_interval = 300: This will detect potential malicious re-org where the attacking fork length will be more than 10 blocks.

PR #809, opened 2019-12-27
Enable minting native tokens according to config

ninjaahhh:
this is the reason why minting doesn't work in testnet3.2


ninjaahhh:
should use "override" field in config https://github.com/QuarkChain/pyquarkchain/blob/ceeea35fa4b87a7c50e4c0363eb5e3a6804a5703/quarkchain/evm/tests/new_statetest_utils.py#L168

@hanyunx will take over this PR

PR #815, opened 2020-02-05
Set up github actions

ninjaahhh:
scaffold, trying to migrate CI off circleci to github actions

### DIFF
--- a/.github/workflows/python-build-and-test.yml
+++ b/.github/workflows/python-build-and-test.yml
@@ -1,0 +1,18 @@
+name: Python unit tests
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/python-build-and-test.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: actions/checkout@v2
+    - name: Build qkchash
+      run: |
+        cd qkchash && make clean && make && cd -
+    - name: Install Project Dependencies
+      run: |
+        pip3 install -e .
+    - name: Test with pytest
+      run: |
+        python3 -m pytest quarkchain/cluster/tests/test_cluster.py
```
qizhou: Should we run all tests or just test_cluster.py?

FreshmanQ: This is just a template, I'm finishing the rest tests

PR #820, opened 2020-02-10
Update CI badge in README

ninjaahhh:


ninjaahhh:
according to [this post](https://github.community/t5/GitHub-Actions/Parallelism-in-Self-Hosted-Runners/td-p/39158), self-hosted runner can only run one job at a time :(

PR #825, opened 2020-02-21
Add GRPC Slave and send them root blocks

### LINKED ISSUE #812: Add gRPC client to master server / root chain logic
...which will tell customized shard servers what minor block has been confirmed

FreshmanQ:
Based on the original cluster_config file, set the first two slaves still be QKC type, the rest two are LIBRA type.

resolves #812


quarkchain/cluster/cluster_config.py
```diff
@@ -92,6 +92,7 @@ class SlaveConfig(BaseConfig):
     WEBSOCKET_JSON_RPC_PORT = None
     ID = ""
     CHAIN_MASK_LIST = None
+    TYPE = ""
```
ninjaahhh: can set the default type here instead of the constructor

quarkchain/cluster/rpc.py
```diff
@@ -76,13 +76,15 @@ class SlaveInfo(Serializable):
         ("host", PrependedSizeBytesSerializer(4)),
         ("port", uint16),
         ("chain_mask_list", PrependedSizeListSerializer(4, ChainMask)),
+        ("type", PrependedSizeBytesSerializer(4)),
```
ninjaahhh: we probably don't want to have the type information for now. reasons is `SlaveInfo` RPC is used to establish full connections between master/slaves, however at our first step we don't have full mechanism to do this with libra slave

so I suggest when establishing connections, skip slaves if their type is not the default.

quarkchain/cluster/slave.py
```diff
@@ -683,7 +684,15 @@ def close_connection(self, conn):
...
     def __init__(
-        self, env, reader, writer, slave_server, slave_id, chain_mask_list, name=None
+        self,
+        env,
+        reader,
+        writer,
+        slave_server,
+        slave_id,
+        chain_mask_list,
+        type,
```
ninjaahhh: same reason as before. `slave.py` is for QKC slaves, we don't need them to be aware of other types of slaves

quarkchain/cluster/cluster_config.py
```diff
@@ -415,7 +419,10 @@ def __create_from_args_internal():
                 slave_config.PORT = args.port_start + i
                 slave_config.ID = "S{}".format(i)
                 slave_config.CHAIN_MASK_LIST = [ChainMask(i | args.num_slaves)]
-
+                if i < 2:
```
ninjaahhh: what does this mean?

FreshmanQ: since we have 4 slaves, just for testing, the first 2 slaves are still QKC, the rest are LIBRA

quarkchain/cluster/cluster_config.py
FreshmanQ: done

quarkchain/cluster/cluster_config.py
ninjaahhh: but you do know slave number is configurable, right? it can be 8, 16, or whatever number

quarkchain/cluster/rpc.py
FreshmanQ: if I skip connection them, then __has_all_shards() checks failed, the cluster will be shut down.

quarkchain/cluster/cluster_config.py
FreshmanQ: do you mean I can set random id for libra slave?

quarkchain/cluster/rpc.py
ninjaahhh: then you should have "normal" slaves covering all shards, but have separate additional slaves to make grpc calls

quarkchain/cluster/cluster_config.py
ninjaahhh: I mean your logic in line 422 will break if in production we have different number of slaves in config. 

I would suggest in command-line flags, don't allow setting types for slave servers

FreshmanQ: done

quarkchain/cluster/rpc.py
FreshmanQ: done

quarkchain/cluster/cluster_config.py
```diff
@@ -351,9 +352,9 @@ def attach_arguments(cls, parser):
 
         parser.add_argument("--monitoring_kafka_rest_address", default="", type=str)
 
-        parser.add_argument(
-            "--libra_slaves", default=ClusterConfig.LIBRA_SLAVES, type=str
-        )
+    #        parser.add_argument(
```
ninjaahhh: delete

quarkchain/cluster/rpc.py
ninjaahhh: not needed?

quarkchain/cluster/rpc.py
```diff
@@ -76,13 +76,15 @@ class SlaveInfo(Serializable):
...
     ]
 
-    def __init__(self, id, host, port, chain_mask_list):
+    def __init__(self, id, host, port, chain_mask_list, type):
```
ninjaahhh: same. seems not needed

quarkchain/cluster/cluster_config.py
```diff
@@ -427,8 +426,10 @@ def __create_from_args_internal():
             with open(args.cluster_config) as f:
                 config = cls.from_json(f.read())
                 config.json_filepath = args.cluster_config
+
```
hanyunx: remove the empty line, also recover L459, L877 in last commit?

quarkchain/cluster/cluster_config.py
```diff
@@ -151,21 +151,26 @@ class ClusterConfig(BaseConfig):
     START_SIMULATED_MINING = False
     CLEAN = False
     GENESIS_DIR = None
+
     QUARKCHAIN = None
     MASTER = None
     SLAVE_LIST = None
+    GRPC_LIST = None
```
ninjaahhh: nit: `GRPC_SLAVE_LIST`

quarkchain/cluster/cluster_config.py
```diff
@@ -151,21 +151,26 @@ class ClusterConfig(BaseConfig):
...
     def __init__(self):
         self.QUARKCHAIN = QuarkChainConfig()
         self.MASTER = MasterConfig()
         self.SLAVE_LIST = []  # type: List[SlaveConfig]
+        self.GRPC_LIST = []  # type: List[SlaveConfig]
         self.SIMPLE_NETWORK = SimpleNetworkConfig()
         self._json_filepath = None
         self.MONITORING = MonitoringConfig()
         self.kafka_logger = KafkaSampleLogger(self)
+
```
ninjaahhh: why have so many blank lines?

quarkchain/cluster/cluster_config.py
```diff
@@ -444,6 +444,7 @@ def to_dict(self):
         ret["MONITORING"] = self.MONITORING.to_dict()
         ret["MASTER"] = self.MASTER.to_dict()
         ret["SLAVE_LIST"] = [s.to_dict() for s in self.SLAVE_LIST]
```
ninjaahhh: you may want to exclude GRPC slaves?

quarkchain/cluster/cluster_config.py
```diff
@@ -459,6 +460,7 @@ def from_dict(cls, d):
         config.MONITORING = MonitoringConfig.from_dict(config.MONITORING)
         config.MASTER = MasterConfig.from_dict(config.MASTER)
         config.SLAVE_LIST = [SlaveConfig.from_dict(s) for s in config.SLAVE_LIST]
+        config.GRPC_LIST = [s for s in config.SLAVE_LIST if s.TYPE == "GRPC"]
```
ninjaahhh: same as before

quarkchain/cluster/master.py
```diff
@@ -1297,6 +1287,10 @@ def handle_new_root_block_header(self, header, peer):
         future_list = self.broadcast_rpc(
             op=ClusterOp.ADD_ROOT_BLOCK_REQUEST, req=AddRootBlockRequest(r_block, False)
         )
+
+        for _ in self.cluster_config.GRPC_LIST:
+            GrpcClient().set_rootchain_confirmed_block()
```
ninjaahhh: you sure you want to use default parameters? every slave may have different host / port, right?

quarkchain/cluster/cluster_config.py
FreshmanQ: same as original format

quarkchain/cluster/master.py
```diff
@@ -1286,6 +1287,10 @@ def handle_new_root_block_header(self, header, peer):
         future_list = self.broadcast_rpc(
             op=ClusterOp.ADD_ROOT_BLOCK_REQUEST, req=AddRootBlockRequest(r_block, False)
         )
+
+        for grpc_slave in self.cluster_config.GRPC_SLAVE_LIST:
+            GrpcClient(grpc_slave.HOST, grpc_slave.PORT).set_rootchain_confirmed_block()
```
ninjaahhh: we don't want to initialize a grpc client every time when we want to send a request. should have the clients initialized before.

quarkchain/cluster/cluster_config.py
```diff
@@ -440,7 +443,10 @@ def to_dict(self):
         ret["QUARKCHAIN"] = self.QUARKCHAIN.to_dict()
         ret["MONITORING"] = self.MONITORING.to_dict()
         ret["MASTER"] = self.MASTER.to_dict()
-        ret["SLAVE_LIST"] = [s.to_dict() for s in self.SLAVE_LIST]
+        ret["SLAVE_LIST"] = [s.to_dict() for s in self.SLAVE_LIST if s.TYPE == "QKCRPC"]
```
ninjaahhh: recommend iterating as follows

```
for s in self.SLAVE_LIST:
  if s.TYPE == "QKCRPC": xxx
  elif s.TYPE == "GRPC": yyy
  else: raise ValueError("unrecognized slave type: %s" % s.TYPE)
```

quarkchain/cluster/cluster_config.py
```diff
@@ -455,7 +461,12 @@ def from_dict(cls, d):
         config.QUARKCHAIN = QuarkChainConfig.from_dict(config.QUARKCHAIN)
         config.MONITORING = MonitoringConfig.from_dict(config.MONITORING)
         config.MASTER = MasterConfig.from_dict(config.MASTER)
-        config.SLAVE_LIST = [SlaveConfig.from_dict(s) for s in config.SLAVE_LIST]
+        config.SLAVE_LIST = [
+            SlaveConfig.from_dict(s)
+            for s in config.SLAVE_LIST
+            if SlaveConfig.from_dict(s).TYPE == "QKCRPC"
+        ]
+        config.GRPC_SLAVE_LIST = [s for s in config.SLAVE_LIST if s.TYPE == "GRPC"]
```
ninjaahhh: `config.SLAVE_LIST` already filtered, this seems wrong. please add a unit test for the config to make sure

quarkchain/cluster/cluster_config.py
```diff
@@ -455,7 +458,22 @@ def from_dict(cls, d):
         config.QUARKCHAIN = QuarkChainConfig.from_dict(config.QUARKCHAIN)
         config.MONITORING = MonitoringConfig.from_dict(config.MONITORING)
         config.MASTER = MasterConfig.from_dict(config.MASTER)
-        config.SLAVE_LIST = [SlaveConfig.from_dict(s) for s in config.SLAVE_LIST]
+        temp_GRPC_SLAVE_LIST = []
```
ninjaahhh: at this step, should already have `config.SLAVE_LIST` as a list of dicts with type `QKCRPC` and `config.GRPC_SLAVE_LIST` as a list of dicts with type `GRPC`, right? do we need to filter again?

quarkchain/cluster/master.py
```diff
@@ -1286,6 +1293,11 @@ def handle_new_root_block_header(self, header, peer):
         future_list = self.broadcast_rpc(
             op=ClusterOp.ADD_ROOT_BLOCK_REQUEST, req=AddRootBlockRequest(r_block, False)
         )
+
+        for grpc_client in self.grpc_slave_pool:
+            result = grpc_client.set_rootchain_confirmed_block()
+            self.grpc_result.append(result)
```
ninjaahhh: I don't think we need persist the results to `self`

quarkchain/cluster/cluster_config.py
FreshmanQ: this step is to solve future type transformation loading from command line as we discussed before. if I remove it, GRPC_LIST will be load as dict rather than object. However, if we just think about loading from config file, then we don't need this step.

quarkchain/cluster/master.py
FreshmanQ: it's just used for testing since we don't have return in this function

quarkchain/cluster/cluster_config.py
ninjaahhh: why not just 

```
config.SLAVE_LIST = [SlaveConfig.from_dict(s) for s in config.SLAVE_LIST]
config.GRPC_SLAVE_LIST = [SlaveConfig.from_dict(s) for s in config.GRPC_SLAVE_LIST]
```

quarkchain/cluster/master.py
ninjaahhh: you should find another way to test the results of the call, instead of storing them into `Master`

quarkchain/cluster/tests/test_utils.py
```diff
@@ -317,12 +317,14 @@ def create_test_clusters(
     chain_size,
     shard_size,
     num_slaves,
+    num_grpc_slaves,
```
ninjaahhh: also make this optional? or use the same number as `num_slaves`?

quarkchain/cluster/tests/test_utils.py
```diff
@@ -372,6 +374,15 @@ def create_test_clusters(
             slave_config.CHAIN_MASK_LIST = [ChainMask(num_slaves | j)]
             env.cluster_config.SLAVE_LIST.append(slave_config)
 
+        if connect_grpc:
+            env.cluster_config.GRPC_SLAVE_LIST = []
+            for j in range(num_grpc_slaves):
+                grpc_slave_config = SlaveConfig()
+                grpc_slave_config.ID = "GS{}".format(j)
+                grpc_slave_config.HOST = "localhost"
+                grpc_slave_config.PORT = 50051 + j
```
ninjaahhh: should probably use `get_next_port`?

quarkchain/cluster/master.py
```diff
@@ -891,6 +893,10 @@ def __has_all_shards(self):
                 if slave.has_shard(full_shard_id):
                     self.branch_to_slaves.setdefault(full_shard_id, []).append(slave)
 
+        for grpc_slave in self.cluster_config.GRPC_SLAVE_LIST:
```
ninjaahhh: this may be moved to constructor since it doesn't depend on normal slaves.

also, one-liner:

```python
self.grpc_slave_pool = [GrpcClient(s.HOST, s.PORT) for s in self.cluster_config.GRPC_SLAVE_LIST]
```

quarkchain/cluster/master.py
```diff
@@ -1286,6 +1292,12 @@ def handle_new_root_block_header(self, header, peer):
         future_list = self.broadcast_rpc(
             op=ClusterOp.ADD_ROOT_BLOCK_REQUEST, req=AddRootBlockRequest(r_block, False)
         )
+
+        for i, grpc_client in enumerate(self.grpc_slave_pool):
+            grpc_client.set_rootchain_confirmed_block(
+                self.cluster_config.GRPC_SLAVE_LIST[i].ID
```
ninjaahhh: adding a TODO saying this is a fake message

ninjaahhh: a second thought. we probably shouldn't use the fake message here, because it needs to be removed very soon. 

suggest we remove the `message` field in the proto.

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -43,7 +47,21 @@ def _tip_gen(shard_state):
     return b
 
 
+class NormalServer(grpc_pb2_grpc.ClusterSlaveServicer):
+    def SetRootChainConfirmedBlock(self, request, context):
+        unittest.TestCase.assertTrue(self, expr=request.message)
```
ninjaahhh: you can do the following

```
class NormalServer()..

  def __init__(self):
    self.request_num = 0
```

then increment the counter in the handling method. then confirm the request number is correct

quarkchain/cluster/master.py
```diff
@@ -1286,6 +1290,10 @@ def handle_new_root_block_header(self, header, peer):
         future_list = self.broadcast_rpc(
             op=ClusterOp.ADD_ROOT_BLOCK_REQUEST, req=AddRootBlockRequest(r_block, False)
         )
+
+        for i, grpc_client in enumerate(self.grpc_slave_pool):
```
ninjaahhh: you don't need enumerate then

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -43,7 +47,24 @@ def _tip_gen(shard_state):
     return b
 
 
+class NormalServer(grpc_pb2_grpc.ClusterSlaveServicer):
```
ninjaahhh: nit: rename to `MockGrpcServer`

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -2430,3 +2451,91 @@ def mock_get_root_chain_stakes(recipient, _):
...
+                clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST[0].PORT,
+            )
+            server1.start()
+            server2, port_number2 = self.build_test_server(
+                normalserver1,
+                clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST[1].HOST,
+                clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST[1].PORT,
+            )
+            server2.start()
+            self.assertEqual(
```
ninjaahhh: I don't think it's necessary to assert `port_number` since they are not part of the code we want to test

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -2430,3 +2451,91 @@ def mock_get_root_chain_stakes(recipient, _):
             # fail again, because quota used up
             with self.assertRaises(ValueError):
                 add_root_block(staker_addr, sign=True)
+
+    def test_grpc_slave(self):
+        id1 = Identity.create_random_identity()
+        acc1 = Address.create_from_identity(id1, full_shard_key=0)
+
+        with ClusterContext(1, acc1, connect_grpc=True) as clusters:
+            block_header_list = [clusters[0].get_shard_state(2 | 0).header_tip]
```
ninjaahhh: we don't really need this setup. a simple root block should be enough. can remove line from 2460 to 2481 using empty minor block header list

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -2430,3 +2451,91 @@ def mock_get_root_chain_stakes(recipient, _):
...
+            self.assertEqual(
+                normalserver1.request_num,
+                len(clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST),
+            )
+            server1.stop(0)
+            server2.stop(0)
+
+            # Test Case 2 ###################################################
+            # This case tests the connection with wrong port number
+            normalserver2 = NormalServer()
```
ninjaahhh: this test case doesn't add much value

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -2457,85 +2457,26 @@ def test_grpc_slave(self):
...
-
-            root_block = clusters[0].master.root_state.create_block_to_mine(
-                block_header_list, acc1
-            )
             # Test Case 1 ###################################################
             # This case tests the correct connection
-            normalserver1 = NormalServer()
-            server1, port_number1 = self.build_test_server(
+            root_block = clusters[0].master.root_state.create_block_to_mine([])
+            normalserver1 = MockGrpcServer()
```
ninjaahhh: nit: rename `normalserver1` to `server`, `server1` / `server2` to `grpc_server1` / `grpc_server2`

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -2457,85 +2457,26 @@ def test_grpc_slave(self):
...
+            server1 = self.build_test_server(
                 normalserver1,
                 clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST[0].HOST,
```
ninjaahhh: can save some characters if you have a separate variable.
```
grpc_slaves = clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST
```

PR #829, opened 2020-02-26
Root block conversion to proto

lyhe18:
Add simplified libra root block structure to grpc_client.proto. The draft structure of libra root block mainly includes the root block id, block height, hashes, signature and minor block headers. More adjustments of proto and grpc_client.py will be made accordingly.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,12 @@
+all: proto qkchash
+
+proto: quarkchain/cluster/grpc_client.proto
+	python -m grpc_tools.protoc --proto_path=quarkchain/cluster --python_out=quarkchain/generated --grpc_python_out=quarkchain/generated grpc_client.proto
+
+.PHONY: qkchash
+qkchash:
+	$(MAKE) -C qkchash
+
+clean:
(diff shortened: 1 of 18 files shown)


makefile
```diff
@@ -0,0 +1,12 @@
+all: proto qkchash
```
ninjaahhh: have you merged master? many files seem to be outdated in this PR

quarkchain/cluster/grpc_client.proto
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
```
ninjaahhh: should be using the one defined in another repo as source of truth (once the latest PR has been merged)

makefile
ninjaahhh: this should be removed

quarkchain/cluster/grpc_client_test.py
```diff
@@ -0,0 +1,88 @@
+import unittest
```
ninjaahhh: this should also be removed?

PR #839, opened 2020-03-09
Call grpc client

lyhe18:
Add call to methods of grpc client in add_root_block() function in master.py.  These methods are responsible for setting confirmed root blocks and broadcasting these root blocks to all shards. Tests are also added in test_cluster.py.

### DIFF
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -51,34 +51,6 @@ class GrpcClient:
             return False
 
         if response.status.code == 0:
-            if str(root_block.minor_block_header_list) == str(
-                request.minor_block_headers
-            ):
-                return True
-            else:
-                return False
+            return True
         else:
             return False
-
-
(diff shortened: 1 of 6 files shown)


quarkchain/cluster/grpc_client.py
```diff
@@ -2,6 +2,7 @@
 import argparse
 import grpc
 from quarkchain.generated import grpc_pb2, grpc_pb2_grpc
+from quarkchain.core import RootBlock, RootBlockHeader, MinorBlockHeader
```
FreshmanQ: RootBlockHeader & MinorBlockHeader is never used, u can remove them

quarkchain/cluster/tests/test_grpc_client.py
```diff
@@ -2,29 +2,65 @@
 import unittest
 
 from concurrent import futures
+
+from typing import List
```
FreshmanQ: this is not needed

PR #842, opened 2020-03-09
Support AddRootBlock GRPC

lyhe18:
(Open this new pull request to prevent from picking up unnecessary histories.)
In this pr, we add call to methods of GRPC client in add_root_block() function of master.py. This method is to  broadcast root blocks to all shards. Tests are also added in test_cluster.py.
For network failure, we first considered different types of errors such as connection lost and timeout error. However, under these conditions, we are unable to retry to connect GRPC slave again, nor could we get a timeout error if we don't specify a timeout for the operations.  As a result, we simply conclude all failures as network error and application error and describe what kind of failure it is with an error message.

### DIFF
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -51,12 +51,12 @@ class GrpcClient:
             Logger.log_exception()
             return False
 
-        if response.status.code == 0:
-            return True
-        else:
-            Logger.warning(
+        if response.status.code != 0:
+            Logger.error(
                 "GRPC failure: {}, {}".format(
                     response.status.code, response.status.message
                 )
             )
             return False
+
+        return True
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -3,6 +3,7 @@ import argparse
 import grpc
 from quarkchain.generated import grpc_pb2, grpc_pb2_grpc
 from quarkchain.core import RootBlock, RootBlockHeader, MinorBlockHeader
+from quarkchain.utils import Logger
 
 HOST = "localhost"
 PORT = 50051
@@ -46,10 +47,16 @@ class GrpcClient:
         )
         try:
             response = self.client.AddRootBlock(request)
-        except Exception:
+        except grpc.RpcError as e:
+            Logger.log_exception()
             return False
 
         if response.status.code == 0:
             return True
         else:
+            Logger.warning(
+                "GRPC failure: {}, {}".format(
+                    response.status.code, response.status.message
+                )
+            )
             return False
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -1293,6 +1293,7 @@ class MasterServer:
 
         for grpc_client in self.grpc_slave_pool:
             grpc_client.set_rootchain_confirmed_block()
+            grpc_client.add_root_block(r_block)
 
         result_list = await asyncio.gather(*future_list)
         check(all([resp.error_code == 0 for _, resp, _ in result_list]))
--- a/quarkchain/cluster/master.py
+++ b/quarkchain/cluster/master.py
@@ -1292,7 +1292,6 @@ class MasterServer:
         )
 
         for grpc_client in self.grpc_slave_pool:
-            grpc_client.set_rootchain_confirmed_block()
             grpc_client.add_root_block(r_block)
 
         result_list = await asyncio.gather(*future_list)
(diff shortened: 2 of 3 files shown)


quarkchain/cluster/master.py
```diff
@@ -1293,6 +1293,7 @@ def handle_new_root_block_header(self, header, peer):
 
         for grpc_client in self.grpc_slave_pool:
             grpc_client.set_rootchain_confirmed_block()
+            grpc_client.add_root_block(r_block)
```
ninjaahhh: can remove lien 1295

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -48,15 +48,32 @@ def _tip_gen(shard_state):
...
+        )  # type: List[MinorBlockHeader]
 
     def SetRootChainConfirmedBlock(self, request, context):
         self.request_num += 1
         return grpc_pb2.SetRootChainConfirmedBlockResponse(
             status=grpc_pb2.ClusterSlaveStatus(code=0, message="Test")
         )
 
+    def AddRootBlock(self, request, context):
+        assert len(self.expected_minor_block_headers) == len(
```
ninjaahhh: this test should be different with the one testing grpc client: we don't care about what `grpc_client.add_root_block` does, we only care about whether it's been called or not. so I suggest just using `request_num` to record number of invocations

quarkchain/cluster/master.py
lyhe18: Ok, then where should I add this method?

ninjaahhh: no need to. it's not being used right now

lyhe18: ok I got it.

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -48,15 +48,20 @@ def _tip_gen(shard_state):
 
 
 class MockGrpcServer(grpc_pb2_grpc.ClusterSlaveServicer):
-    def __init__(self):
+    def __init__(self, expected_minor_block_headers):
```
ninjaahhh: not needed?

quarkchain/cluster/tests/test_cluster.py
```diff
@@ -2461,7 +2466,7 @@ def test_grpc_slave(self):
             # This case tests the correct connection
             root_block = clusters[0].master.root_state.create_block_to_mine([])
             grpc_slaves = clusters[0].master.env.cluster_config.GRPC_SLAVE_LIST
-            server = MockGrpcServer()
+            server = MockGrpcServer(root_block.minor_block_header_list)
```
ninjaahhh: same

quarkchain/cluster/tests/test_cluster.py
lyhe18: Done.

quarkchain/cluster/tests/test_cluster.py
lyhe18: Removed.

quarkchain/cluster/tests/test_cluster.py
lyhe18: Removed.

quarkchain/cluster/master.py
ninjaahhh: now we need to think about failure cases: what if the result doesn't succeed? should we handle failures differently for network failures (e.g. timeout) and application failures (e.g. the slave thinks the root block is malformed)? is Boolean type enough for this RPC? what do you think?

lyhe18: Do we need to add more failure types in GrpcClient.add_root_block()? Maybe we can define a GrpcResponse struct in rpc.py and use different error code representing different failure types.
For the root block sent to the slave, do we need to first validate the content of it before we send it to the slave?

quarkchain/cluster/grpc_client.py
```diff
@@ -46,10 +47,16 @@ def add_root_block(self, root_block: RootBlock) -> bool:
...
             response = self.client.AddRootBlock(request)
-        except Exception:
+        except grpc.RpcError as e:
+            Logger.log_exception()
             return False
 
         if response.status.code == 0:
             return True
         else:
+            Logger.warning(
```
ninjaahhh: consider `Logger.error`

also, a nit suggestion on coding styles (inspired from golang) that is we can handle errors first, then proceed to normal case. so I would suggest following:

```
if response.status.code != 0:
  # warning and return False

return True
```

lyhe18: Changed.

PR #854, opened 2020-03-23
Update gas for transfer mnt precompiled

ninjaahhh:
gas calculation should be similar to how CALL is being handled: intrinsic gas + value transfer cost

### DIFF
--- a/quarkchain/evm/specials.py
+++ b/quarkchain/evm/specials.py
@@ -241,7 +241,7 @@ def proc_transfer_mnt(ext, msg):
     if msg.data.size < 96 or msg.static:
         return 0, 0, []
     to = utils.int_to_addr(msg.data.extract32(0))
-    mnt = msg.data.extract32(32)
+    token_id = msg.data.extract32(32)
     value = msg.data.extract32(64)
     data = msg.data.extract_all(96)
 
@@ -261,7 +261,10 @@ def proc_transfer_mnt(ext, msg):
     if msg.gas < gascost:
         return 0, 0, []
     # Handle insufficient balance or exceeding max call depth
-    if ext.get_balance(msg.sender, token_id=mnt) < value or msg.depth >= vm.MAX_DEPTH:
+    if (
+        ext.get_balance(msg.sender, token_id=token_id) < value
+        or msg.depth >= vm.MAX_DEPTH
+    ):
         return 0, msg.gas - gascost, []
     new_msg = vm.Message(
         msg.sender,
@@ -272,7 +275,7 @@ def proc_transfer_mnt(ext, msg):
         msg.depth + 1,
         code_address=to,
         static=False,
-        transfer_token_id=mnt,
+        transfer_token_id=token_id,
         gas_token_id=msg.gas_token_id,
     )
     return apply_msg(ext, new_msg)
--- a/quarkchain/evm/specials.py
+++ b/quarkchain/evm/specials.py
@@ -237,16 +237,32 @@ def proc_current_mnt_id(ext, msg):
 def proc_transfer_mnt(ext, msg):
     from quarkchain.evm.messages import apply_msg
 
-    # Data must be >= 96 bytes
-    if msg.data.size < 96:
-        return 0, 0, []
-    gascost = 3
-    if msg.gas < gascost:
+    # Data must be >= 96 bytes and static call not allowed
+    if msg.data.size < 96 or msg.static:
         return 0, 0, []
     to = utils.int_to_addr(msg.data.extract32(0))
     mnt = msg.data.extract32(32)
     value = msg.data.extract32(64)
     data = msg.data.extract_all(96)
+
+    num_zero_bytes = data.count(ascii_chr(0))
+    num_non_zero_bytes = len(data) - num_zero_bytes
+    # Intrinsic gas
+    gascost = (
+        opcodes.GTXCOST
+        + opcodes.GTXDATAZERO * num_zero_bytes
+        + opcodes.GTXDATANONZERO * num_non_zero_bytes
+    )
+    if value > 0:
+        gascost += opcodes.GCALLVALUETRANSFER
+        if not ext.account_exists(to):
+            gascost += opcodes.GCALLNEWACCOUNT
+    # Out of gas
+    if msg.gas < gascost:
+        return 0, 0, []
+    # Handle insufficient balance or exceeding max call depth
+    if ext.get_balance(msg.sender, token_id=mnt) < value or msg.depth >= vm.MAX_DEPTH:
+        return 0, msg.gas - gascost, []
     new_msg = vm.Message(
         msg.sender,
         to,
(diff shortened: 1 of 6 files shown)


quarkchain/evm/specials.py
```diff
@@ -237,16 +237,32 @@ def proc_current_mnt_id(ext, msg):
...
         return 0, 0, []
     to = utils.int_to_addr(msg.data.extract32(0))
     mnt = msg.data.extract32(32)
     value = msg.data.extract32(64)
     data = msg.data.extract_all(96)
+
+    num_zero_bytes = data.count(ascii_chr(0))
+    num_non_zero_bytes = len(data) - num_zero_bytes
+    # Intrinsic gas
+    gascost = (
```
ninjaahhh: a second thought: as the caller already pays the intrinsic gas for the tx itself (including the origin data), should we add an another layer of intrinsic cost? @qizhou

qizhou: I think we don't need to pay the intrinsic gas here - intrinsic gas should be only used when the tx is initiated, but I haven't seen the intrinsic gas about a sub-message call.  Could you check it again?

quarkchain/evm/specials.py
```diff
@@ -237,16 +237,32 @@ def proc_current_mnt_id(ext, msg):
...
+    )
+    if value > 0:
+        gascost += opcodes.GCALLVALUETRANSFER
+        if not ext.account_exists(to):
+            gascost += opcodes.GCALLNEWACCOUNT
+    # Out of gas
+    if msg.gas < gascost:
+        return 0, 0, []
+    # Handle insufficient balance or exceeding max call depth
+    if ext.get_balance(msg.sender, token_id=mnt) < value or msg.depth >= vm.MAX_DEPTH:
```
qizhou: I think we may still add GSTIPENT in submsg gas (assuming GCALLVALUETRANSFER > GSTIPENT) since GSTIPENT may not be supplied when calling the contract.

Further, we may not allow "to" to be any special contracts to eliminate the reentrancy issue (or other possible corner cases).

quarkchain/evm/specials.py
```diff
@@ -237,16 +237,32 @@ def proc_current_mnt_id(ext, msg):
...
-    # Data must be >= 96 bytes
-    if msg.data.size < 96:
-        return 0, 0, []
-    gascost = 3
-    if msg.gas < gascost:
+    # Data must be >= 96 bytes and static call not allowed
+    if msg.data.size < 96 or msg.static:
         return 0, 0, []
     to = utils.int_to_addr(msg.data.extract32(0))
     mnt = msg.data.extract32(32)
```
qizhou: Looks like the mnt can be greater than uint128.  Should we have any test to covert this?

quarkchain/evm/specials.py
```diff
@@ -237,16 +237,32 @@ def proc_current_mnt_id(ext, msg):
 def proc_transfer_mnt(ext, msg):
     from quarkchain.evm.messages import apply_msg
 
-    # Data must be >= 96 bytes
-    if msg.data.size < 96:
-        return 0, 0, []
-    gascost = 3
-    if msg.gas < gascost:
+    # Data must be >= 96 bytes and static call not allowed
+    if msg.data.size < 96 or msg.static:
```
qizhou: I think it is right to check msg.static, but do you think we need to accept an input parameter to issue a static call?  It looks like the static call should be used for all transfer/send in solidity rather than supplying gas with GSTIPENT

qizhou: Oh, I double-check the code again, and actually, to call STATICCALL, the value must be 0 in EVM.  This means we should still reject any call with msg.static = true.

quarkchain/evm/specials.py
ninjaahhh: good catch. we have set `TOKEN_ID_MAX` to 4873763662273663091, and should check for all other methods `proc_transfer_mnt` and `proc_mint_mnt`

quarkchain/evm/specials.py
ninjaahhh: users can set `to` as another address which in turn calls a special contract. we control this kind of recursive invocation

ninjaahhh: 1. sounds good
1. seems enough to limit `to` not to be itself, otherwise if we limit `to` not be **any** special address, we have to deal with potential conflicts in the future if we add more special addresses.

quarkchain/evm/specials.py
ninjaahhh: done

quarkchain/evm/tests/test_specials.py
```diff
@@ -99,3 +106,59 @@ def test_proc_balance_mnt(self):
...
+        self.assertEqual(balance, 30000)
+
+        new_addr = b"\x01" * 20
+        testcases = [
+            # Format: (index, description, to, value, gas, expected ret, post check)
+            (1, "gas not enough", sender, 0, 699, (0, 0, []), None),
+            (2, "no value transfer with min gas", sender, 0, 700, (1, 0, []), None),
+            (3, "value transfer needs more gas", sender, 1, 9699, (0, 0, []), None),
+            # Should have stipend gas left
+            (4, "value transfer needs more gas", sender, 1, 9700, (1, 2300, []), None),
```
ninjaahhh: @qizhou: this is how stipend would work, minimum gas cost met (GCALL 700 + GVALUETRANSFER 9000 = 9700), while still having 2300 left (GSTIPEND), which essentially lowers the cost of GVALUETRANSFER from 9000 to 6700

qizhou: Great!  That is much clearer!

quarkchain/evm/specials.py
```diff
@@ -237,26 +237,46 @@ def proc_current_mnt_id(ext, msg):
...
+    token_id = msg.data.extract32(32)
     value = msg.data.extract32(64)
     data = msg.data.extract_all(96)
+
+    # Token ID should be within range
+    if token_id > TOKEN_ID_MAX:
+        return 0, 0, []
+
+    # Doesn't allow target address to be this precompiled contract itself
+    if to == decode_hex(b"000000000000000000000000000000514b430002"):
```
qizhou: b"000000000000000000000000000000514b430002" is already in bytes - should we do decode_hex() again?

quarkchain/evm/specials.py
```diff
@@ -237,26 +237,46 @@ def proc_current_mnt_id(ext, msg):
...
+        return 0, 0, []
+
+    gascost = 700  # gascost of CALL
```
qizhou: I think the 700 gascost is already paid by the caller?  The essential goal is that if we implement a CALLX with the same functionality as precompiled contracts, and the cost should be close.

quarkchain/evm/specials.py
ninjaahhh: yes, since b"0000..." is not the real byte representation

```python
In [1]: from quarkchain.evm.specials import *

In [2]: decode_hex(b"000000000000000000000000000000514b430002")
Out[2]: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00QKC\x00\x02'
```

quarkchain/evm/specials.py
ninjaahhh: i see

ninjaahhh: done

PR #855, opened 2020-03-24
adding tx to queue needs to check gasprice in QKC

qizhou:
This diff will enforce the minimum gasprice in QKC using converted QKC gasprice instead of gasprice of a native token directly.  This avoids that a tx of a non-native token may be discarded since its gasprice is too small.

Following the fix, a couple of following things should be done:
- More cluster tests with mnt
- The transaction queue will order the txs by their gasprice in QKC instead of gasprice in their native token.

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -39,7 +39,7 @@ from quarkchain.evm.messages import (
     apply_transaction,
     validate_transaction,
     apply_xshard_deposit,
-    get_genesis_gasprice,
+    convert_to_default_chain_token_gasprice,
 )
 from quarkchain.evm.specials import SystemContract
 from quarkchain.evm.state import State as EvmState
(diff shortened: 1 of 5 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -538,6 +539,16 @@ def add_tx(self, tx: TypedTransaction, xshard_gas_limit=None):
             evm_tx = self.__validate_tx(
                 tx, evm_state, xshard_gas_limit=xshard_gas_limit
             )
+
+            # Don't add the tx if the gasprice in QKC is too low.
+            # Note that this is not enforced by consensus,
+            # but miners will likely discard the tx if the gasprice is too low.
+            genesis_gasprice = get_genesis_gasprice(
```
ninjaahhh: nit: rename to `default_token_gasprice` and `get_default_token_gasprice()`??

quarkchain/cluster/shard_state.py
```diff
@@ -1189,13 +1200,15 @@ def __add_transactions_to_block(self, block: MinorBlock, evm_state: EvmState):
                 break
 
             evm_tx = tx.tx.to_evm_tx()
+            evm_tx.set_quark_chain_config(self.env.quark_chain_config)
 
+            genesis_gasprice = get_genesis_gasprice(
```
ninjaahhh: same here

quarkchain/cluster/shard_state.py
qizhou: Looks like we use genesis_token and default_token interchangeably although genesis_token is global-wise while default_token is chain-wise.  So the question is which variable should we stick to?

ninjaahhh: good call. I'm seeing inconsistent usages, such as `apply_transaction` is using genesis token while `validate_transaction` is using default chain token. 

the fundamental question is, if we have a new chain with default chain token set to something else (e.g. QZEC / QBTC), the conversion should apply for them, instead of QKC, right? so I would suggest using default chain token and double check existing usage of genesis token is consistent with our expectation

PR #861, opened 2020-03-26
add uint128_max check for tx gasprice, startgas, token id fields

qizhou:


ninjaahhh:
note this change may lead to hard forks if we had previous tx of big values.

cc @pingke since this is a consensus change

ninjaahhh:
> note this change may lead to hard forks if we had previous tx of big values.
> 
> cc @pingke since this is a consensus change

realized gas limit should be fine because the block gas limit is in the order of 10mil (way less than uint128_max). and if gas price is uint128_max then the total gas fee would be (2**128 * 21000 / 1e18) = 7e25 qkc which is way more than our total supply.

quarkchain/evm/messages.py
```diff
@@ -128,6 +135,15 @@ def validate_transaction(state, tx):
     if not tx.sender:  # sender is set and validated on Transaction initialization
         raise UnsignedTransaction(tx)
 
+    # (1a) startgas, gasprice, gas token id, transfer token id must be <= UINT128_MAX
+    if (
+        tx.startgas > UINT128_MAX
+        or tx.gasprice > UINT128_MAX
+        or tx.gas_token_id > UINT128_MAX
```
ninjaahhh: use `TOKEN_ID_MAX` for token ID?

qizhou:
> > note this change may lead to hard forks if we had previous tx of big values.
> > cc @pingke since this is a consensus change
> 
> realized gas limit should be fine because the block gas limit is in the order of 10mil (way less than uint128_max). and if gas price is uint128_max then the total gas fee would be (2**128 * 21000 / 1e18) = 7e25 qkc which is way more than our total supply.

Yes, this should not result in a consensus change as
1, we only allow token id to be QKC at the moment
2, QKC total supply is far below uint128_max
3, startgas will be limited by block gas limit

However, if a native token with large supply (we allow up to uint256_max) and extremely high exchange rate, then calling contracts with overflowed values may happen.

PR #864, opened 2020-03-27
Reduce shard staking requirement for PoSW

ninjaahhh:
as our block reward drop comes near (factor = 0.88), we should also reduce PoSW staking requirement accordingly on shards.

### DIFF
--- a/quarkchain/cluster/root_state.py
+++ b/quarkchain/cluster/root_state.py
@@ -665,4 +665,6 @@ class RootState:
             block.header.hash_prev_block,
             self.db.get_root_block_header_by_hash,
         )
-        return get_posw_info(config, block.header, lambda: stakes, block_cnt, signer)
+        return get_posw_info(
+            config, block.header, lambda: stakes, block_cnt, signer=signer
+        )
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -1843,7 +1843,7 @@ class ShardState:
                 self.env.quark_chain_config.genesis_token,
             ),
             block_cnt,
-            stake_per_block,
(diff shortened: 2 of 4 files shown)


ninjaahhh:
this would be a consensus change cc @pingke @qizhou 

also the targeting shard height for block reward reduction is 3153600, as of now (10:38am 3/30 PT) we are at around 2950000, so roughly about 28 days to go

quarkchain/cluster/root_state.py
```diff
@@ -665,4 +665,6 @@ def get_posw_info(
             block.header.hash_prev_block,
             self.db.get_root_block_header_by_hash,
         )
-        return get_posw_info(config, block.header, lambda: stakes, block_cnt, signer)
+        return get_posw_info(
+            config, block.header, lambda: stakes, block_cnt, signer=signer
```
hanyunx: what is this change for?

ninjaahhh: `get_posw_info` now has one more optional parameter, if `signer` is not specified, it would be regarded for the added `total_stakes_per_block` parameter

hanyunx: Got it, thanks!

quarkchain/cluster/shard_state.py
```diff
@@ -1848,6 +1843,7 @@ def _posw_info(self, block: MinorBlock) -> Optional[PoSWInfo]:
                 self.env.quark_chain_config.genesis_token,
             ),
             block_cnt,
+            stake_per_block,
```
qizhou: I would suggest using a named argument (stake_per_block=stake_per_block) since the definition of the function assumes it is named.

qizhou:
image

Observed reduced staking

PR #867, opened 2020-03-30
Add GRPC req/resp log

lyhe18:
Logs are added to record the length of unconfirmed headers, the successfully sent root blocks and received minor blocks. This pr is correlated with https://github.com/QuarkChain/libra-sharded/pull/56

### DIFF
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -46,7 +46,7 @@ class GrpcClient:
             minor_block_headers=minor_block_header_list,
         )
         Logger.info(
-            "Sent root block id: {}, height: {}, length of included minor block headers: {}".format(
+            "Root block sent, id: {}, height: {}, length of included minor block headers: {}".format(
                 request.id.hex(), request.height, len(request.minor_block_headers)
             )
         )
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -45,6 +45,14 @@ class GrpcClient:
             height=height,
             minor_block_headers=minor_block_header_list,
         )
+        Logger.info(
+            "Sent root block: {}, height: {}".format(request.id.hex(), request.height)
+        )
+        Logger.info(
+            "Length of included minor block headers: {}".format(
+                len(request.minor_block_headers)
+            )
+        )
         try:
             response = self.client.AddRootBlock(request)
         except grpc.RpcError as e:
@@ -58,11 +66,6 @@ class GrpcClient:
                 )
             )
             return False
-        Logger.info(
-            "Length of included minor block headers: {}".format(
-                len(request.minor_block_headers)
-            )
-        )
         return True
 
     def get_unconfirmed_header(self) -> Optional[List[Any]]:
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -68,5 +68,7 @@ class GrpcClient:
         except grpc.RpcError:
             Logger.log_exception()
             return None
-
+        Logger.info(
+            "Length of unconfirmed headers: {}".format(len(response.header_list))
+        )
         return response.header_list
--- a/quarkchain/cluster/grpc_client.py
+++ b/quarkchain/cluster/grpc_client.py
@@ -46,7 +46,7 @@ class GrpcClient:
             minor_block_headers=minor_block_header_list,
         )
         Logger.info(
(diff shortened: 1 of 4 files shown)


quarkchain/cluster/master.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, root_state):
 
     def AddMinorBlockHeader(self, request, context):
         self.root_state.add_validated_minor_block_hash(request.id, {})
+        # Logger.info("Minor block received: {}".format(request.id))
```
FreshmanQ: remove comments

ninjaahhh:
you have the same problem as in another PR: this PR should only be about logging, you don't want to include other irrelevant changes

FreshmanQ:
this PR should only be related to log info. can you branch out from dev and re-submitting this PR?

quarkchain/cluster/root_state.py
```diff
@@ -281,6 +281,7 @@ def get_tip_block(self):
         return self.db.get_root_block_by_hash(self.tip.get_hash())
 
     def add_validated_minor_block_hash(self, hash: bytes, coinbase_tokens: Dict):
+        Logger.info("Minor block received: {}".format(hash.hex()))
```
ninjaahhh: instead of logging inside root state, I think we should log at the grpc response handling place, because we don't want to affect normal non-grpc setup

quarkchain/cluster/grpc_client.py
```diff
@@ -68,5 +68,7 @@ def get_unconfirmed_header(self) -> Optional[List[Any]]:
         except grpc.RpcError:
             Logger.log_exception()
             return None
-
+        Logger.info(
+            "Length of unconfirmed headers: {}".format(len(response.header_list))
```
ninjaahhh: also log number of minor block headers in the root block when calling `AddRootBlock`?

lyhe18: Ok~

quarkchain/cluster/grpc_client.py
```diff
@@ -58,7 +58,11 @@ def add_root_block(self, root_block: RootBlock) -> bool:
                 )
             )
             return False
-
+        Logger.info(
+            "Length of included minor block headers: {}".format(
```
ninjaahhh: also log root block height and hash?

lyhe18: Ok~~

quarkchain/cluster/grpc_client.py
```diff
@@ -45,6 +45,14 @@ def add_root_block(self, root_block: RootBlock) -> bool:
             height=height,
             minor_block_headers=minor_block_header_list,
         )
+        Logger.info(
```
ninjaahhh: can you put those 2 logs together?

lyhe18: No prob~

quarkchain/cluster/grpc_client.py
```diff
@@ -45,6 +45,11 @@ def add_root_block(self, root_block: RootBlock) -> bool:
...
+            "Sent root block id: {}, height: {}, Length of included minor block headers: {}".format(
```
ninjaahhh: nit: `Length` -> `length`

quarkchain/cluster/miner.py
```diff
@@ -432,15 +432,14 @@ def _log_status(block: Block):
...
+        Logger.info(
             "[{}] {} [{}] ({:.2f}) {}".format(
                 full_shard_id,
                 block.header.height,
                 count,
                 elapsed,
                 block.header.get_hash().hex(),
             ),
-            60,
         )
```
FreshmanQ: we have to set 60 back here since we have already log root block in AddRootBlock request

ninjaahhh: good point

lyhe18: Got it~

quarkchain/cluster/grpc_client.py
```diff
@@ -45,6 +45,11 @@ def add_root_block(self, root_block: RootBlock) -> bool:
             height=height,
             minor_block_headers=minor_block_header_list,
         )
+        Logger.info(
+            "Sent root block id: {}, height: {}, length of included minor block headers: {}".format(
+                request.id.hex(), request.height, len(request.minor_block_headers)
```
FreshmanQ: prefer log format consists with Libra root block height, id, header_list

PR #870, opened 2020-04-01
setup MNT system contract deploy timestamp

qizhou:
Enabled at Friday, May 1, 2020 12:00:00 AM GMT.


ninjaahhh:
according to our config parsing, suggest configuring those timestamps in our JSON config file

https://github.com/QuarkChain/pyquarkchain/blob/00ab12e4d839ce78eb20465db71356822fab4ecc/quarkchain/env.py#L49-L69

qizhou:
> according to our config parsing, suggest configuring those timestamps in our JSON config file
> 
> https://github.com/QuarkChain/pyquarkchain/blob/00ab12e4d839ce78eb20465db71356822fab4ecc/quarkchain/env.py#L49-L69

Updated when it is mainnet, the default timestamps will be automatically loaded.

PR #872, opened 2020-04-03
add chain_id restriction for token mint and system contract deployment

qizhou:


quarkchain/evm/specials.py
```diff
@@ -426,6 +447,13 @@ def addr(self) -> bytes:
     ),
 }
 
+SYSTEM_CONTRACT_SCOPE_MAP = {
+    # Should be LOCAL_CHAIN_0.  Use GLOBAL to maintain backward compatibility
+    SystemContract.ROOT_CHAIN_POSW: SystemContractScope.GLOBAL,
```
ninjaahhh, line 452: how about just putting the scope variables into the `_system_contracts` in line 432 dictionary as part of the value, instead of another dict lookup?

qizhou: That was the plan, but it seems that it will change all the code that uses _system_contracts, while the map will impact a few lines of code.  Since the time is limited, I think using the map would be a simple and fast fix.

ninjaahhh:
cc @516108736

PR #877, opened 2020-04-07
fix equal cmp for TokenBalanceMap

qizhou:
If the token map has a balance zero, reconstructing it from db will result in an empty map, which differs from a map with zero balance, and thus equal comparison will fail.

This fix uses a serialized version of the token map to compare, which could be implemented in a more efficient way, but it is sufficient at the moment (where the token balance map is mostly a single item).

### DIFF
--- a/quarkchain/core.py
+++ b/quarkchain/core.py
@@ -587,6 +587,10 @@ class TokenBalanceMap(Serializable):
     def get_hash(self):
         return sha3_256(self.serialize())
 
+    def __eq__(self, other):
+        # TODO: Could implement in a more-efficient way, but make sure skip_func works.
+        return self.serialize() == other.serialize()
+
 
 def calculate_merkle_root(item_list):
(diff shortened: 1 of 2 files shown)


ninjaahhh:
where do we use equality comparison with token maps?

qizhou:
When we compare two headers - one is serialized from db, and another is created in-flight, and if there is a zero balance, such comparison will fail.

Example code is here:
https://github.com/QuarkChain/pyquarkchain/blob/00ab12e4d839ce78eb20465db71356822fab4ecc/quarkchain/cluster/shard_state.py#L1155

qizhou:
That will override dict class, which may change several dict semantics.  Another way I think is to remove 0 balance when calling __init__, but I am not sure this will affect existing code that may access balance_map directly.

> I see. this seems a quick fix. maybe we should find a way to force token balance map automatically gets rid of zero-value entries like overriding `__setitem__` method?

PR #880, opened 2020-04-08
Set default MIN_GAS_PRICE in testnet 3.3

FreshmanQ:


testnet/3.3/cluster_config_template.json
```diff
@@ -11,6 +11,8 @@
   "CLEAN": false,
   "GENESIS_DIR": null,
   "QUARKCHAIN": {
+    "MIN_TX_POOL_GAS_PRICE": 0,
```
qizhou, line 14: Should we set the value so that we could test the feature?

FreshmanD: yes, if we don't set, it won't same with converted gas price.

qizhou: But setting it as 0 will essentially include all tx no matter the gas price is?

The feature means the "only include txs with at least the gas price in QKC"

FreshmanQ: During test 2 ** 128 (3.4e38) overflow, since the default `MIN_TX_POOL_GAS_PRICE` is 10 ** 9, if we set ratio lower than 3.4e29, it will return false. So does this mean we have to restrain users don't set ratio too low?

qizhou: Yes, if the ratio is too low, then such tokens will not be able to be used as a gas token.

PR #882, opened 2020-04-13
update contract bytecode to da899a29

qizhou:


quarkchain/constants.py
```diff
@@ -35,16 +35,16 @@
 
 # The following code is compiled with:
 #    solc (0.5.17+commit.d19bba13.Linux.g++) --bin --optimize --optimize-runs 200 --evm-version petersburg
-# Source code commit 896b2f38.
+# Source code commit da899a29.
 NON_RESERVED_NATIVE_TOKEN_CONTRACT_BYTECODE = bytes.fromhex(
     """
```
ninjaahhh, line 40: thought you are going to use a different address for supervisor?

qizhou: Will change to another address later.

PR #887, opened 2020-04-22
cluster.py prints cluster config in env

qizhou:
Previous cluster.py only prints raw cluster config, which doesn't reflect the actual cluster config that applies mainnet default configs (such as stake reduction ts, smart contract deployment).

This version will print the env.config, which is the one used by master and slaves.


qizhou:
Test plan:
1, Remove ts
image
2, ./run_cluster.sh and check whether ts's are applied
image

PR #892, opened 2020-06-30
Scaffold total balance count RPC

ninjaahhh:
also as a code lab for building RPC calls inside the cluster.

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -1987,3 +1987,12 @@ class ShardState:
             self.env.quark_chain_config.block_reward_decay_factor.denominator ** epoch
         )
         return value * decay_numerator // decay_denominator
+
+    def get_total_balance(
+        self,
+        token_id: int,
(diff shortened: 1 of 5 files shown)


quarkchain/cluster/shard_state.py
```diff
@@ -1987,3 +1987,12 @@ def __decay_by_epoch(self, value: int, block_height: int):
...
         return value * decay_numerator // decay_denominator
+
+    def get_total_balance(
+        self,
+        token_id: int,
+        block_hash: bytes,
+        limit: int,
+        starter: Optional[bytes] = None,
+    ) -> Tuple[int, bytes]:
+        return 42, b""
```
qizhou, line 1998: It may worth commenting that whether starter is inclusive or exclusive, and when EOF is reached, what would the value of the returned next be.

ninjaahhh: good call. but it also depends on how the underlying trie implements the iteration. once we have some grasp on it we will update the comments.

PR #894, opened 2020-07-07
Increase weekly checkdb timeout

ninjaahhh:
recently we are seeing timeouts for weekly check-db job which runs through the whole dataset (like this [run](https://github.com/QuarkChain/pyquarkchain/runs/838287334)). this PR increases the timeout from 1 day to 2 days


qizhou:
Interesting findings are:
- 17 days ago, it took 14h
image

- 24 days ago, it took 12h
image

Wondering why it took such a long time these days?

PR #896, opened 2020-07-14
Add batch size limitation

pichaoqkc:
Add batch size limitation and rename the JSON parameter.

### DIFF
--- a/quarkchain/cluster/jsonrpc.py
+++ b/quarkchain/cluster/jsonrpc.py
@@ -1281,7 +1281,7 @@ class JSONRPCHttpServer:
     @decode_arg("limit", quantity_decoder)
     async def getTotalBalance(self, block_hash, token_id, address, limit="0x64"):
         if limit > 10000:
-            raise InvalidParams
+            limit = 10000
         starter = Address.create_from(address)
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/jsonrpc.py
```diff
@@ -1280,6 +1280,8 @@ async def setTargetBlockTime(self, root_block_time=0, minor_block_time=0):
     @decode_arg("address", address_decoder)
     @decode_arg("limit", quantity_decoder)
     async def getTotalBalance(self, block_hash, token_id, address, limit="0x64"):
+        if limit > 10000:
```
ninjaahhh, line 1283: how about following `getTransactionsByAddress`
image

pichaoqkc: Well, if we change the limit silently, the user cannot know how many accounts did the returned balances stand for, since we didn't return the changed limit.

ninjaahhh: but it's a `limit`, not a `size` parameter. we don't have the API contract that says the returned balance is guaranteed to be for that many addresses, but only for pagination purpose

pichaoqkc: Make sense

PR #898, opened 2020-07-15
Improve getTotalBalance JSONRPC

ninjaahhh:
...update for better ergonomics where recipient can be none and full
shard key is included in block ID. a sample query will be like:

```bash
curl -s -X POST -d  '{"jsonrpc": "2.0", "method": "getTotalBalance", "params": ["0x1a1debffe5e5293467f1677d857343058916bab82caa661f8087ddcf4f3fcad400000001"], "id": 1}' localhost:38391
```

note default value for token is QKC, recipient is 0x0 and limit is 100

### DIFF
--- a/quarkchain/cluster/jsonrpc.py
+++ b/quarkchain/cluster/jsonrpc.py
@@ -1299,9 +1299,7 @@ class JSONRPCHttpServer:
         total_balance, next_starter = result
         return {
             "totalBalance": quantity_encoder(total_balance),
-            "next": data_encoder(
-                next_starter + full_shard_key.to_bytes(4, byteorder="big")
-            ),
+            "next": data_encoder(next_starter),
         }
 
     @private_methods.add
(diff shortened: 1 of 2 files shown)


quarkchain/cluster/jsonrpc.py
```diff
@@ -1299,9 +1299,7 @@ async def getTotalBalance(
         total_balance, next_starter = result
         return {
             "totalBalance": quantity_encoder(total_balance),
-            "next": data_encoder(
-                next_starter + full_shard_key.to_bytes(4, byteorder="big")
-            ),
+            "next": data_encoder(next_starter),
```
pichaoqkc, line 1302: Now the next address cannot be reused directly?

ninjaahhh: it's a recipient, the func parameter is also a recipient

PR #893, opened 2020-07-07
Implement total balance counting in shards

lyhe18:

PR #901, opened 2020-07-15
Fix total balance counting off-by-one

ninjaahhh:
#893 has a non-obvious optimization which can reduce number of iterations but kinda unintuitive. 

suppose we only have 1 active address,  starting `get_total_balance` with `None` as the `starter` and 1 as `limit`, how many calls should we have? in my opinion, it should go like

1. first run, returns `(bal, <active address>)`
2. second run using that active address as starter (note it's exclusive on `starter`), returns `(0, 0x0)`

however previous code in #893 only needs one run returning `(bal, 0x0)` which essentially just looks ahead and skips that address when counting the total balance. 

it's not wrong, but I prefer the more straightforward logic.

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -2020,24 +2020,16 @@ class ShardState:
         """
         evm_state = self._get_evm_state_from_hash(block_hash)
         if not evm_state:
-            # TBD, may do hash validation before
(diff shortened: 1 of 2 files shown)


ninjaahhh:
also removed the constraint that starter must be an existing address. seems unnecessary

PR #904, opened 2020-07-20
Update bytecode with 60s overtime duration

hanyunx:
for automatic testing


qizhou:
Will this break replayability of existing mainnet data?

ninjaahhh:
> Will this break replayability of existing mainnet data?

it's for devnet branch

PR #905, opened 2020-07-21
Fix typing import missing in prometheus client

pichaoqkc:
import typing


ninjaahhh:
make sure your code works before merging.

also, please have a more explicit PR title

PR #908, opened 2020-07-24
Fix balance count if not all shards are in one root block

ninjaahhh:
will try continuously fetching root blocks until all shards' latest minor block hashes are acquired.

can't do `query_balance.py` style because other processes (e.g. `cluster.py`) may already be using db.

### DIFF
--- a/quarkchain/tools/count_total_balance.py
+++ b/quarkchain/tools/count_total_balance.py
@@ -1,7 +1,7 @@
 import argparse
 import functools
 import logging
-from typing import List, Tuple
+from typing import List, Tuple, Dict, Any
 
 import jsonrpcclient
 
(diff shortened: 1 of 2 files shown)


quarkchain/tools/count_total_balance.py
```diff
@@ -14,44 +14,61 @@
 logging.getLogger("jsonrpcclient.client.response").setLevel(logging.WARNING)
 
 TIMEOUT = 10
-
-host = "http://localhost:38391"
+TOTAL_SHARD = 8
```
pichaoqkc, line 17: Do we have a more elegant way to get the current total_shard?

ninjaahhh: `getNetworkInfo` JSONRPC API should do, but I think for now this is fine for our network

PR #909, opened 2020-07-25
Update balance count prometheus gauge

ninjaahhh:
we shouldn't be using labels to record numeric values, instead, we can have a separate gauge for block height, thus if anomaly happens we can trace back to the root block height causing the problem

### DIFF
--- a/quarkchain/cluster/prom.py
+++ b/quarkchain/cluster/prom.py
@@ -87,9 +87,9 @@ def prometheus_balance(args):
             # Rpc not ready, wait and try again.
             time.sleep(3)
             continue
+        block_height_gauge.set(latest_block_height)
         for token_name, (_, balance_dict) in total_balance.items():
             balance_gauge.labels("total", token_name).set(sum(balance_dict.values()))
-            block_height_gauge.set(latest_block_height)
             for shard_id, shard_bal in balance_dict.items():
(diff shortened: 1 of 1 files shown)


quarkchain/cluster/prom.py
```diff
@@ -86,14 +87,11 @@ def prometheus_balance(args):
             # Rpc not ready, wait and try again.
             time.sleep(3)
             continue
-        for token_name, (time_stamp, balance_dict) in total_balance.items():
-            balance_gauge.labels(
-                latest_block_height, time_stamp, "total", token_name
-            ).set(sum(balance_dict.values()))
+        for token_name, (_, balance_dict) in total_balance.items():
+            balance_gauge.labels("total", token_name).set(sum(balance_dict.values()))
+            block_height_gauge.set(latest_block_height)
```
pichaoqkc, line 92: In this way, you cannot map a specific height to the balance it has, but assume that block may have the balance with the same time series fetched by Prometheus?

ninjaahhh: correct. we are missing balances for certain heights. our current code doesn't promise all blocks are queried

pichaoqkc: Yes, our work is more like "sampling" the block height and get their balance, so we are actually performing a "fake" real-time monitoring. In this way, there is no need to match them one by one

PR #913, opened 2020-08-04
Add prev_headers to EVM state 

ninjaahhh:
~WIP~ tests to be added

### DIFF
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -350,10 +350,12 @@ class ShardState:
             state.timestamp = timestamp
         # iterate until reaches genesis or header list reaches 256
         # since most headers are in LRU cache, this should not affect performance too much
-        while block_hash and len(state.prev_headers) < 256:
-            state.prev_headers.append(block_hash)
-            prev = self.db.get_minor_block_header_by_hash(block_hash)
-            block_hash = prev and prev.hash_prev_minor_block
+        while len(state.prev_headers) < 256:
+            h = self.db.get_minor_block_header_by_hash(block_hash)
+            if not h:
+                break
+            state.prev_headers.append(h)
+            block_hash = h.hash_prev_minor_block
         return state
 
     def init_genesis_state(self, root_block):
--- a/quarkchain/cluster/shard_state.py
+++ b/quarkchain/cluster/shard_state.py
@@ -394,7 +394,9 @@ class ShardState:
         self.header_tip = genesis_block.header
         self.meta_tip = genesis_block.meta
         self.evm_state = self.__create_evm_state(
-            genesis_block.meta.hash_evm_state_root, sender_disallow_map={}
+            genesis_block.meta.hash_evm_state_root,
+            sender_disallow_map={},
+            block_hash=genesis_block.header.get_hash(),
         )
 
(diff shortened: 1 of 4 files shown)


ninjaahhh:
used following contract for testing and tested with goquarkchain on devnet, syncing works as expected.

```
pragma solidity >=0.4.22 <0.6.0;

pragma experimental ABIEncoderV2;

contract RecordBlockHash {
    
    mapping (uint256 => bytes32) public hashes;

    function record(uint256 start) external {
        for (uint256 i = 0; i <= 200; i++) {
            uint256 n = block.number - (start + i);
            bytes32 h = blockhash(n);
            hashes[n] = h;
        }
    }
}
```

ninjaahhh:
still running DB check for the whole data. meanwhile, take a look when you have time @qizhou

quarkchain/cluster/shard_state.py
```diff
@@ -774,6 +787,7 @@ def run_block(self, block, evm_state=None, x_shard_receive_tx_list=None):
         for k, v in pure_coinbase_amount.balance_map.items():
             evm_state.delta_token_balance(evm_state.block_coinbase, k, v)
 
+        evm_state.add_block_header(block.header)
```
qizhou, line 790: Why do we need to add a block header here?

quarkchain/cluster/shard_state.py
```diff
@@ -344,6 +348,14 @@ def __create_evm_state(
         state.sender_disallow_map = sender_disallow_map
         if timestamp:
             state.timestamp = timestamp
+        # iterate until reaches genesis or header list reaches 256
+        # since most headers are in LRU cache, this should not affect performance too much
+        while block_hash and len(state.prev_headers) < 256:
+            h = self.db.get_minor_block_header_by_hash(block_hash)
+            if not h:
+                break
+            state.prev_headers.append(h)
```
qizhou, line 357: Why not use state.add_block_header?  Looks like add_block_header appendleft, but this appendright, but I cannot see why we need to appendleft?  Perhaps we could replace appendright to appendleft in add_block_header if appendleft is not used anywhere?

ninjaahhh: note the order of `prev_headers` is descending, i.e. header at height 100, heigh 99, etc. both this part and `evm_state.add_block_header(block.header)` will keep this invariant (`appendleft` is used in `evm_state.add_block_header` since the passed block is usually at the highest height).

quarkchain/cluster/shard_state.py
ninjaahhh: after running the block, evm_state will commit the underlying changes, while we also want to make sure its `prev_header` is up to date, so the next time evm_state executing txs, it will have the latest block header available

ninjaahhh:
> still running DB check for the whole data. meanwhile, take a look when you have time @qizhou

running check DB from height 690366 finishes. now trying to sync to the latest. once confirmed working, will merge this PR

ninjaahhh:
> > still running DB check for the whole data. meanwhile, take a look when you have time @qizhou
> 
> running check DB from height 690366 finishes. now trying to sync to the latest. once confirmed working, will merge this PR

done. syncing now. merging this PR

PR #936, opened 2021-11-12
improve mnt transfer EVM test

ninjaahhh:
1. fixed test util's display of prev & current state
2. updated fixture's contract bytecode and made sure EVM behavior is
   expected


qizhou:
So my understanding is to test large MNT token id?

qizhou (approved):

ninjaahhh:
> So my understanding is to test large MNT token id?

good call, will add as a separate test. this update only fixes the previously problematic test

PR #942, opened 2022-03-18
Update mainnet config for root chain posw

ping-ke:


quarkchain/cluster/shard_state.py
```diff
@@ -1183,26 +1183,32 @@ def get_next_block_coinbase_amount(self):
...
+        Logger.debug("{} blocks to be confirm".format(steps))
+        max_height = start_height + max_blocks
+        st = time_ms()
+        for i in range(steps):
+            if header.height <= max_height:
+                header_list.append(header)
             header = self.db.get_minor_block_header_by_hash(
                 header.hash_prev_minor_block
             )
+        Logger.info("get_unconfirmed_header_list: unconfirmed header {}, max_blocks = {}, time use {} ms."
```
qizhou, line 1203: What is the purpose of logging info here?

mainnet/singularity/cluster_config_template.json
```diff
@@ -90,10 +90,10 @@
         "DIFF_DIVIDER": 10000,
         "WINDOW_SIZE": 512,
         "TOTAL_STAKE_PER_BLOCK": 1000000000000000000000000,
-        "BOOST_TIMESTAMP": 1646064000,
+        "BOOST_TIMESTAMP": 1649044800,
```
qizhou, line 93: Since it took us one week to test devnet, should we extend the timestamp to another week?

quarkchain/cluster/shard_state.py
ping-ke: From this log you can see:
1. Whether there is Header accumulation;
2. If there is accumulation, how much does it affect the performance;
3. Is there any message congestion? (When multiple pieces of this message from the same shard appear at once but have not appeared for some times before, may this message be blocked?)

qizhou (approved):

PR #970, opened 2026-03-15
Part 6 - p2p: replace upnpclient with async-upnp-client for NAT port mapping

ping-ke:
## Summary
  `upnpclient` uses synchronous/blocking I/O and `netifaces` for interface discovery; neither is well-maintained on Python 3.13.

  Rewrite `UPnPService` in `nat.py` to use `async-upnp-client`:
  - Use `async_search()` for UPnP IGD device discovery
  - Use `AiohttpSessionRequester` + `UpnpFactory` for async HTTP requests
  - Remove `ThreadPoolExecutor` blocking path and `netifaces` dependency
  - Expose `discover()` coroutine (previously `add_nat_portmap()`) that
    returns the mapped external IP string
  - Update `p2p_server.py` call site accordingly

  ## Test plan
  - [ ] Node starts and attempts UPnP discovery without errors
  - [ ] NAT mapping is established when a compatible router is present
  - [ ] Graceful handling when no UPnP device is found (returns `None`)


quarkchain/p2p/nat.py
```diff
@@ -58,9 +31,42 @@ def __init__(self, port: int, token: CancelToken = None) -> None:
...
+    async def discover(self) -> Optional[str]:
+        """
+        Discover router and create initial port mapping.
+        Returns external IP if successful.
+        """
+        session = aiohttp.ClientSession()
+        try:
+            await self._discover(session)
+        finally:
+            await session.close()
```
qzhodl: Codex found there is a session lifetime bug here.

`discover()` creates an `aiohttp.ClientSession`, passes it into `AiohttpSessionRequester`, and `_discover()` stores a `service` object created from that requester. But the session is closed immediately after `_discover()` returns, while the same `self._service` is still used later by `_add_port_mapping()`, `_get_external_ip()`, refresh in `_run()`, and `_delete_port_mapping()`.

`async-upnp-client` does not recreate the session automatically. If the underlying `ClientSession` has been closed, subsequent calls fail with `RuntimeError: Session is closed`.

Codex verified this with a minimal repro using `AiohttpSessionRequester` directly: create requester with a session, close the session, then call `async_http_request()` and it raises `RuntimeError: Session is closed`.

So if discovery finds a valid WANIP service, this path is likely to fail as soon as we try to call `AddPortMapping` / `GetExternalIPAddress` on `self._service`.

The minimal repro is
```python
import asyncio
import aiohttp

from async_upnp_client.aiohttp import AiohttpSessionRequester
from async_upnp_client.const import HttpRequest


async def main():
    session = aiohttp.ClientSession()
    requester = AiohttpSessionRequester(session)

    await session.close()

    try:
        await requester.async_http_request(
            HttpRequest("GET", "http://127.0.0.1:1", {}, None)
        )
    except Exception as e:
        print(type(e).__name__)
        print(str(e))
    else:
        print("NO_ERROR")


asyncio.run(main())

```

quarkchain/p2p/nat.py
```diff
@@ -58,9 +31,42 @@ def __init__(self, port: int, token: CancelToken = None) -> None:
...
+
+    async def discover(self) -> Optional[str]:
+        """
+        Discover router and create initial port mapping.
+        Returns external IP if successful.
+        """
+        session = aiohttp.ClientSession()
+        try:
+            await self._discover(session)
+        finally:
```
qzhodl: You may need to run the tests on a network with the router to verify that it works

quarkchain/p2p/nat.py
```diff
@@ -1,48 +1,21 @@
+
+import aiohttp
```
qzhodl, line 1: This PR introduces a new runtime dependency on `async-upnp-client` (and directly imports `aiohttp` in `nat.py`), but the project dependencies still only list `upnpclient` and `netifaces`.

Right now `requirements.txt` / `setup.py` are not updated to install the new dependency set, so a clean environment will fail at import time before this code can run.

It looks like we should add `async-upnp-client` (and ensure `aiohttp` is available through project dependencies), while removing the old UPnP-specific dependencies that are no longer used.

quarkchain/p2p/nat.py
```diff
@@ -72,112 +78,133 @@ async def _run(self) -> None:
...
-                # of software, either way we should not override it.
-                # https://tools.ietf.org/id/draft-ietf-pcp-upnp-igd-interworking-07.html#errors
-                self.logger.info(
-                    "NAT port mapping already configured, not overriding it"
-                )
-            else:
-                self.logger.exception("Failed to setup NAT portmap")
+            await asyncio.wait_for(
+                async_search(on_response),
+                timeout=UPNP_DISCOVER_TIMEOUT_SECONDS,
```
qzhodl: I think the timeout handling here is still not doing what the constant/comment suggests.

`UPNP_DISCOVER_TIMEOUT_SECONDS` is set to 30, but that value is only used in `asyncio.wait_for(...)` around `async_search(on_response)`. The `async-upnp-client` implementation of `async_search()` has its own internal timeout/sleep behavior, and this code does not pass `UPNP_DISCOVER_TIMEOUT_SECONDS` into `async_search()` itself.

So the effective discovery duration here is still governed by the library default rather than the 30-second constant defined in this file. If we want the search duration to actually be 30 seconds, we should pass `timeout=UPNP_DISCOVER_TIMEOUT_SECONDS` to `async_search(...)` directly.

quarkchain/p2p/nat.py
ping-ke: This will be resolved after merge other changes. Test can be run on branch upgrade-py3-13.

qzhodl: I don’t think this is a good practice, as additional bugs may be introduced during the merge process.

Also, have you run the full test suite again on upgrade-py3-13?

ping-ke: All changes will be made on the `upgrade-py3-13` branch. After all tests pass, they will be cherry-picked to the target branch. Once the merge is completed, the full test suite will be executed again. I will also perform a comparison between the `upgrade/base-line` branch and the `upgrade-py3-13` branch to ensure consistency and correctness.

quarkchain/p2p/nat.py
ping-ke: I don't have a UPnP-enabled router environment, so I covered it with unit tests instead. The tests mock async_search, UpnpFactory, and the WANIPConnection service to verify the full lifecycle: discover → add port mapping → refresh via _run → delete mapping → session cleanup. Edge cases (no device, device creation error, non-WANIPConn device, service exception) are also covered.

quarkchain/p2p/nat.py
```diff
@@ -58,9 +31,46 @@ def __init__(self, port: int, token: CancelToken = None) -> None:
...
-        )  # : PortMapping when called externally, this never returns None
+        self._session = None
+        self._service = None
+
+
+    # -----------------------------
+    # Public API
+    # -----------------------------
+
+    async def discover(self) -> Optional[str]:
```
qzhodl, line 45: I think the error-handling behavior here is still riskier than the previous implementation.

Before this refactor, NAT setup was effectively best-effort: `add_nat_portmap()` logged common UPnP failures and returned `None`, so the server could still start even if port mapping failed. In the new flow, `BaseServer._run()` directly awaits `self.upnp_service.discover()`, and `discover()` only handles the “no WANIP service found” case gracefully.

If `_discover()` raises, or if `_add_port_mapping()` fails after a service has already been found, that exception now bubbles out of `discover()` and can abort server startup. In that failure path, the session also does not appear to be closed before the exception escapes.

It would be safer to preserve the previous best-effort behavior here: log the NAT setup failure, clean up the session, and return `None` so startup can continue without UPnP.

quarkchain/p2p/tests/test_nat.py
```diff
@@ -0,0 +1,483 @@
+import asyncio
```
qzhodl, line 1: The added unit tests are useful, but I do not think they are enough on their own for this kind of networking change.

This PR replaces the UPnP implementation, changes how the internal IP is selected, and changes the mapping/session lifecycle. Those paths can pass with mocks while still failing against real routers or real local-network configurations.

Could we update the test results with at least one real-environment/manual validation run against an actual UPnP-capable router? That would make it much easier to assess whether the new implementation behaves correctly outside the mocked unit-test setup.

ping-ke: Run `quarkchain/p2p/nat.py` on `98.47.106.208`

```
python3 -m quarkchain.p2p.nat
Internal IP: 192.168.0.185
External IP (before discover): None

Discovering UPnP devices (timeout 30s)...
I0414 04:33:51.005412 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.006463 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.007445 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.008324 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.009185 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.010130 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.010987 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.011945 nat.py:134] Found UPnP WANIP service
I0414 04:33:51.012848 nat.py:134] Found UPnP WANIP service
I0414 04:34:21.004581 nat.py:145] Adding port mapping 38291->192.168.0.185:38291
External IP: 98.47.106.208
Port 38291 mapped successfully
Press Enter to remove mapping and exit...
I0414 04:37:50.365177 nat.py:202] Deleted UPnP port mapping
Mapping removed
```

quarkchain/p2p/nat.py
```diff
@@ -72,112 +100,161 @@ async def _run(self) -> None:
...
-        except upnpclient.soap.SOAPError as e:
-            if e.args == (718, "ConflictInMappingEntry"):
-                # An entry already exists with the parameters we specified. Maybe the router
-                # didn't clean it up after it expired or it has been configured by other piece
-                # of software, either way we should not override it.
-                # https://tools.ietf.org/id/draft-ietf-pcp-upnp-igd-interworking-07.html#errors
-                self.logger.info(
-                    "NAT port mapping already configured, not overriding it"
+                        raise
+                protocols_added.append(protocol)
```
qzhodl: I think the rollback behavior here is still unsafe in the conflict case.

`UpnpActionResponseError(error_code=718)` is treated above as “the mapping is already in place”, but this protocol is still appended to `protocols_added`. If a later protocol then fails, the rollback path below will call `DeletePortMapping` for this protocol and may remove a pre-existing mapping that we explicitly decided not to override.

For example: if TCP returns 718 and UDP fails, we will roll back TCP even though TCP may have been an existing router entry from a previous run or from other software.

It would be safer to only append to `protocols_added` when we actually created the mapping in this call, and exclude the 718 case from rollback.

quarkchain/p2p/nat.py
```diff
@@ -72,112 +100,161 @@ async def _run(self) -> None:
...
-                continue
-            retv.append(device)
-        return retv
+    def _get_internal_ip(self) -> str:
+        """
+        Robust internal IP detection using socket trick.
+        """
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        try:
+            s.connect(("8.8.8.8", 80))
```
qzhodl: I think this internal-IP selection logic is riskier than the previous implementation.

Before this refactor, we chose the local address by matching the UPnP device’s network, which made the mapping target explicit. Here we infer the address from the default route by “connecting” to `8.8.8.8`, which can pick the wrong interface on multi-homed hosts, VPN setups, or machines with multiple local networks.

In those cases, discovery may succeed against one router while `AddPortMapping` is configured with a different local address, which would be a behavioral regression from the old code.

Could we preserve the old “address on the router’s network” behavior, or otherwise derive the internal IP from the discovered device/location rather than from the default route?

quarkchain/p2p/nat.py
```diff
@@ -58,9 +32,65 @@ def __init__(self, port: int, token: CancelToken = None) -> None:
...
+        Returns external IP if successful, or None if UPnP is unavailable or fails.
+
+        Failures are best-effort: any exception is logged and None is returned so
+        server startup can continue without UPnP.
+
+        Concurrent calls are serialised by _discover_lock so that two callers cannot
+        race on _session / _service state.
+        """
+        async with self._discover_lock:
+            await self._close_session()
```
qzhodl, line 57: I think there is still a stale-state issue across repeated `discover()` calls.

We close the previous session here, but `_service` / `_router_host` are not cleared before the new discovery starts. If `discover()` has already succeeded once, `_discover()` can short-circuit on the old `self._service`, and the rest of this call may continue using a service object that still belongs to the old, already-closed session.

If `discover()` is meant to be repeatable, it would be safer to clear `_service` and `_router_host` before starting a new discovery round.

qzhodl (approved):

quarkchain/p2p/nat.py
```diff
@@ -1,48 +1,22 @@
...
-from concurrent.futures import ThreadPoolExecutor
-import ipaddress
-from typing import List, NamedTuple
+import socket
+from contextlib import suppress
+from typing import Optional
 from urllib.parse import urlparse
 
 from quarkchain.p2p.cancel_token.token import CancelToken, OperationCancelled
-from quarkchain.p2p.exceptions import NoInternalAddressMatchesDevice
```
blockchaindevsh, line 8: Dead code: `NoInternalAddressMatchesDevice` in exceptions.py is no longer used. Clean up in this PR or a follow-up.

quarkchain/p2p/tests/test_nat.py
```diff
@@ -0,0 +1,453 @@
...
+import pytest
+from unittest.mock import MagicMock, AsyncMock, patch
+
+from quarkchain.p2p.cancel_token.token import OperationCancelled
+
+# CancelToken uses deprecated asyncio.Event(loop=loop) removed in py3.10+,
+# patch it to ignore the loop kwarg
+import quarkchain.p2p.cancel_token.token as _ct
+
+_OrigCancelToken = _ct.CancelToken
```
blockchaindevsh: Looks like this patch will become dead code after https://github.com/QuarkChain/pyquarkchain/pull/972 is merged, right?

blockchaindevsh (approved):

PR #971, opened 2026-03-15
Part 5 - jsonrpc: replace jsonrpcserver/jsonrpcclient with custom implementation

ping-ke:
## Summary
  `jsonrpcserver` 3.x and `jsonrpcclient` 2.x are no longer maintained
  and are incompatible with Python 3.13.

  - Add `quarkchain/cluster/jsonrpcserver.py`: a lightweight JSON-RPC 2.0
    server dispatcher built on aiohttp (`RpcMethods`, `JsonRpcError` hierarchy,
    positional and named parameter dispatch)
  - Add `quarkchain/jsonrpc_client.py`: `JsonRpcClient` (synchronous, httpx-based)
    and `AsyncJsonRpcClient` (async, aiohttp-based) replacing the old clients
  - Update `jsonrpc.py` to use the new `RpcMethods` class and make
    `start_*_server()` classmethods async
  - Replace `jsonrpcclient` calls in all tools and `prom.py` with `JsonRpcClient.call()`
  - Rename `quarkchain/tools/stats` → `quarkchain/tools/stats.py` (was missing extension)
  - Update `test_jsonrpc.py` to use `AsyncJsonRpcClient`

  ## Test plan
  - [x] `pytest quarkchain/cluster/tests/test_jsonrpc.py` passes
  - [x] Public and private JSON-RPC endpoints respond correctly
  - [x] `external_miner.py` can fetch work and submit results


qzhodl:
### Code review

Found 2 issues:

1. **Typo: `client.client()` instead of `client.call()` in `stats.py`** -- `basic()` calls `client.client("getStats")` but `JsonRpcClient.client` is the internal `httpx.Client` instance, not a callable RPC method. This will raise `TypeError` at runtime. The `stats()` function on line 35 correctly uses `client.call("getStats")`.

https://github.com/QuarkChain/pyquarkchain/blob/f867cb332b6ff81912b53e8c0041353bebd67d1f/quarkchain/tools/stats.py#L20-L22

2. **`asyncio.get_running_loop()` called from synchronous `__init__`** -- Both `JSONRPCHttpServer.__init__` (line 509) and `JSONRPCWebsocketServer.__init__` (line 1462) call `asyncio.get_running_loop()`. These constructors are invoked from synchronous `main()` in `master.py` and `slave.py` before the event loop is running, which will raise `RuntimeError: no running event loop`. The previous `asyncio.get_event_loop()` works in both sync and async contexts.

https://github.com/QuarkChain/pyquarkchain/blob/f867cb332b6ff81912b53e8c0041353bebd67d1f/quarkchain/cluster/jsonrpc.py#L505-L510

https://github.com/QuarkChain/pyquarkchain/blob/f867cb332b6ff81912b53e8c0041353bebd67d1f/quarkchain/cluster/jsonrpc.py#L1458-L1463

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

quarkchain/cluster/jsonrpc.py
```diff
@@ -542,12 +535,12 @@ async def __handle(self, request):
             self.counters[method] = 1
         # Use armor to prevent the handler from being cancelled when
         # aiohttp server loses connection to client
-        response = await armor(self.handlers.dispatch(request))
+        response = await self.handlers.dispatch(d)
```
qzhodl: The comment says armor prevents handler cancellation—why was it removed?

quarkchain/jsonrpc_client.py
```diff
@@ -0,0 +1,80 @@
...
+import uuid
+
+class JsonRpcError(Exception):
+    def __init__(self, error):
+        self.code = error.get("code")
+        self.message = error.get("message")
+        self.data = error.get("data")
+        super().__init__(f"JSON-RPC Error {self.code}: {self.message}")
+
+class JsonRpcClient:
```
qzhodl, line 16: Why jsonrpc_client.py and jsonrpcserver.py have different naming conventions?

quarkchain/tools/stats.py
```diff
@@ -23,8 +17,8 @@ def fstr(v: float):
     return "{:.2f}".format(v)
 
 
-def basic(client, ip):
-    s = client.send(jsonrpcclient.Request("getStats"))
+def basic(client: JsonRpcClient, ip):
+    s = client.client("getStats")
```
qzhodl: Have you tested the tools after this change?

quarkchain/cluster/tests/test_jsonrpc.py
```diff
@@ -57,14 +47,10 @@ def jrpc_http_server_context(master):
...
-    async def __send_request(*args):
-        async with aiohttp.ClientSession(loop=asyncio.get_event_loop()) as session:
-            client = aiohttpClient(session, "http://localhost:38391")
-            response = await client.request(*args)
-            return response
+rpc_client = AsyncJsonRpcClient("http://localhost:38391")
 
-    return call_async(__send_request(*args))
+def send_request(method, *args):
+     return call_async(rpc_client.call_with_dict_params(method, *args))
```
qzhodl: Do we still have "call_async" function?

ping-ke: This will be replace by await after merge.

quarkchain/jsonrpc_client.py
ping-ke: Rename jsonrpcserver to jsonrpc_server

quarkchain/cluster/jsonrpc.py
```diff
@@ -506,7 +506,7 @@ def start_test_server(cls, env, master_server):
     def __init__(
         self, env, master_server: MasterServer, port, host, methods: RpcMethods
     ):
-        self.loop = asyncio.get_running_loop()
+        self.loop = asyncio.get_event_loop()
```
qzhodl, line 509: Since this change was already made in PR #972, I don’t think we need to revert it?

quarkchain/cluster/jsonrpc.py
ping-ke: Use asyncio.shield to prevent the handler from being cancelled when aiohttp server loses connection to client
```
        response = await asyncio.shield(self.handlers.dispatch(d))
```

quarkchain/cluster/jsonrpc.py
ping-ke: This will be overwrite after merge that PR.

quarkchain/tools/stats.py
ping-ke: I only tested a few of them before, mainly to see if the modifications would work. Now all updated tools have been tested.

quarkchain/cluster/jsonrpc.py
qzhodl: A better way to handle multiple sequential PRs is to branch each subsequent PR off the previous one. This way, earlier changes are automatically included in later PRs, which helps avoid confusion during review and keeps the code easier to manage. It also makes testing more straightforward.

qzhodl (approved):

quarkchain/jsonrpc_client.py
```diff
@@ -0,0 +1,80 @@
...
+        resp = await self.client.post("", json=payload)
+        resp.raise_for_status()
+        data = resp.json()
+
+        if "error" in data:
+            raise JsonRpcError(data["error"])
+
+        return data.get("result")
+    
+    async def call_with_dict_params(self, method, params):
```
syntrust: What's the purpose of this method? Why is there no 'call_with_dict_params' in the sync version?

quarkchain/jsonrpc_client.py
```diff
@@ -0,0 +1,80 @@
...
+            "params": list(params),
+            "id": str(uuid.uuid4()),
+        }
+
+        resp = self.client.post("", json=payload)
+        resp.raise_for_status()
+        data = resp.json()
+
+        if "error" in data:
+            raise RuntimeError(data["error"])
```
syntrust: Is it better to use `JsonRpcError` than `RuntimeError`?

quarkchain/cluster/jsonrpc_server.py
```diff
@@ -0,0 +1,169 @@
...
+                    "message": str(e),
+                },
+                "id": req_id,
+            }
+        except Exception:
+            return {
+                "jsonrpc": "2.0",
+                "error": {
+                    "code": -32603,
+                    "message": "Internal error",
```
syntrust, line 159: Would it be better to at least log the error?

quarkchain/jsonrpc_client.py
ping-ke: The old codebase used the jsonrpcclient library which handled both positional ([a, b]) and named ({k: v}) JSON-RPC params internally. When we replaced it with a hand-rolled client for the py3.13 upgrade, call_with_dict_params was added as a quick way to support named params (used only by createTransactions). It's now been removed — call() handles both formats: pass a list for positional params, pass a dict for named params. The sync version doesn't need this because it's only used in quarkchain/tools where all calls use simple positional params.

syntrust (approved):

quarkchain/jsonrpc_client.py
```diff
@@ -0,0 +1,74 @@
...
+
+    async def call(self, method, *params):
+        # JSON-RPC "params" can be a list (positional) or dict (named).
+        # The old jsonrpcclient library handled this internally; since we
+        # replaced it with a hand-rolled client we replicate the logic here:
+        #   call("method", [a, b])   -> params = [a, b]       (positional)
+        #   call("method", {k: v})   -> params = {k: v}       (named)
+        #   call("method", a, b)     -> params = [a, b]       (positional)
+        if len(params) == 1 and isinstance(params[0], (dict, list)):
+            rpc_params = params[0]
```
blockchaindevsh, line 51: JsonRpcClient (sync) doesn't unwrap single-element params, AsyncJsonRpcClient does. This inconsistency may cause bugs if sync/async clients are used interchangeably.

quarkchain/cluster/jsonrpc_server.py
```diff
@@ -0,0 +1,173 @@
...
+        except JsonRpcError as e:
+            return {
+                "jsonrpc": "2.0",
+                "error": e.to_dict(),
+                "id": req_id,
+            }
+
+        except TypeError as e:
+            # Could be missing/extra arguments → treat as invalid params
+            return {
```
blockchaindevsh: If the handler itself raises a TypeError (e.g., a bug in the handler code like accessing None.something), it's reported as "invalid params" instead of "internal error". This masks real bugs.

quarkchain/cluster/jsonrpc_server.py
```diff
@@ -0,0 +1,173 @@
...
+                "jsonrpc": "2.0",
+                "error": {
+                    "code": -32603,
+                    "message": "Internal error",
+                },
+                "id": req_id,
+            }
+
+    async def aiohttp_handler(self, request: web.Request) -> web.Response:
```
blockchaindevsh: The aiohttp_handler method is never used — the server uses self.handlers.dispatch(d) directly.

blockchaindevsh (approved):

PR #973, opened 2026-03-15
Part 3 - pow: remove pyethash C extension, always use pure Python ethash

ping-ke:
## Summary

`pyethash` is a C++ extension that is not compatible with Python 3.13. The pure Python implementation in `ethereum.pow.ethash` can serve as a replacement; however, directly removing `pyethash` caused a **significant regression in synchronization performance (10–20× slower)**, making block sync impractically slow.

This PR:
- Fixes the compatibility issue by removing `pyethash`
- Introduces a **4-round optimization pipeline (R1–R4)** to recover and exceed the original performance
- Reduces sync time from **25.24s → 0.86s (~29× improvement vs old Python, faster than pyethash)**

---

## Problem

`pyethash 0.1.27` crashes with **segfault or floating point exception** on Python 3.13 when calling `hashimoto_light()`.

Removing `pyethash` and falling back to the existing pure Python implementation leads to:
- Heavy overhead in hex encoding/decoding
- Excessive Python object allocations in hot loops
- **Severe performance degradation in PoW and block synchronization**

---

## Root Cause

A bug in `src/python/core.c` of `pyethash`:

`PyArg_ParseTuple` uses `"y#"` format which writes `Py_ssize_t` (8 bytes on 64-bit) into `int` variables (4 bytes), causing **stack corruption**.

```c
// core.c line 76-77 (pyethash 0.1.27)
int cache_size, header_size;  // BUG: should be Py_ssize_t
if (!PyArg_ParseTuple(args, "k" PY_STRING_FORMAT PY_STRING_FORMAT "K",
    &block_number, &cache_bytes, &cache_size, &header, &header_size, &nonce))
```

## Solution

This PR removes `pyethash` entirely and replaces it with a **progressively optimized Ethash implementation**, eliminating Python bottlenecks in the PoW hot path while restoring (and exceeding) the original performance.

### Optimization Strategy

To systematically eliminate bottlenecks, we applied four incremental optimization rounds, each targeting a different layer:

- **R1–R2 (Python-level optimizations)**  
  Remove serialization overhead and reduce Python object allocations

- **R3 (Cython hot loop)**  
  Move the hottest FNV mixing loop into C

- **R4 (Full C pipeline)**  
  Eliminate Python overhead entirely in the PoW critical path (including Keccak)

**See https://github.com/QuarkChain/pyquarkchain/issues/976 for more detail**

### Root Block Sync Time (end-to-end)

> Syncing **one root block with 144 miniblocks (maximum load)**

impl | sync time | vs pyethash | vs old | speedup vs R2
-- | -- | -- | -- | --
pyethash | 1.47 s | 1× | ~17× | —
old | 25.24 s | ~17× | 1× | —
R1 | 12.38 s | ~8.4× | ~2.0× | —
R2 | 8.52 s | ~5.8× | ~3.0× | 1×
R3 | 1.39 s | ~0.95× | ~18× | ~6×
R4 | 0.86 s | ~0.58× | ~29× | ~10×

- Restores performance after removing `pyethash`
- Achieves **~29× speedup vs original Python implementation**
- Achieves **better performance than pyethash baseline**

## Test plan 
- [x] Mining and PoW verification still pass in existing tests
- [x] No ImportError on Python 3.13
- [x] Benchmarks cover old / R1 / R2 / R3 / R4
- [x] Sync tested end-to-end with profiling and timing logs


ethereum/pow/ethpow.py
```diff
@@ -7,16 +7,9 @@
...
-    import pyethash
-
-    ETHASH_LIB = "pyethash"  # the C++ based implementation
-except ImportError:
-    ETHASH_LIB = "ethash"
-    warnings.warn("using pure python implementation", ImportWarning)
-
 
 # always have python implementation declared
+@lru_cache(10)
```
qzhodl: The @lru_cache(10) on get_cache_slow appears to be redundant. The inner function _get_cache in ethereum/pow/ethash.py already has @lru_cache(10) and caches by (seed, n), which is epoch-based. 

The outer cache on get_cache_slow keys on (cache_size, block_number) — since block_number changes every block, it will mostly miss and provide no real benefit

ethereum/pow/ethpow.py
```diff
@@ -30,33 +23,9 @@ def hashimoto_slow(
 ):
     return ethash.hashimoto_light(full_size, cache, mining_hash, bin_nonce)
 
+get_cache = get_cache_slow
+hashimoto = hashimoto_slow
 
-if ETHASH_LIB == "ethash":
```
qzhodl, line 34: Is there a performance benchmark comparing pyethash.hashimoto_light (C++) vs ethash.hashimoto_light (pure Python)? It would be helpful to see how much slower the pure Python implementation is per call.

ping-ke: python -m ethereum.pow.tests.bench_hashimoto 2
- Pure Python: 19.7s / 100 calls (~197 ms/call)
- pyethash (C++): 0.196s / 100 calls (~2 ms/call)
- ~100x speedup
- This is run on wsl
 
pyethash works on GCC 11.4 (WSL), but fails on:
GCC 13.x (AX101)
GCC 14.x (Docker)

Open issue https://github.com/QuarkChain/pyquarkchain/issues/976 to check whether we want to add back pyethash.

ethereum/pow/tests/bench_hashimoto.py
```diff
@@ -0,0 +1,58 @@
...
+    # pyethash.mkcache_bytes takes block_number directly (mirrors calculate_cache in ethpow.py)
+    cpp_cache = pyethash.mkcache_bytes(BLOCK_NUMBER)
+    nonce_int = big_endian_to_int(NONCE)
+
+    def bench_cpp():
+        # Signature matches ethpow.py line 53-54:
+        #   pyethash.hashimoto_light(block_number, cache_bytes, mining_hash, nonce_int)
+        pyethash.hashimoto_light(BLOCK_NUMBER, cpp_cache, HEADER, nonce_int)
+
+    elapsed_cpp = timeit.timeit(bench_cpp, number=ROUNDS)
```
qzhodl, line 54: It seems that a different cache size is used between cpp and python version

qzhodl (approved):

ethereum/pow/ethpow.py
syntrust: dead import

ping-ke:
Add Performance improvements and related tests/bench. See https://github.com/QuarkChain/pyquarkchain/issues/976 for more infor.

qzhodl:
I didn't see the rust code in the latest change here

ethereum/pow/ethash_utils.py
```diff
@@ -39,73 +26,18 @@ def _sha3_512(x):
...
-
-def deserialize_cache(ds):
-    return [
-        deserialize_hash(ds[i : i + HASH_BYTES]) for i in range(0, len(ds), HASH_BYTES)
-    ]
+def ethash_sha3_512(x: Union[bytes, np.ndarray]) -> np.ndarray:
+    """sha3-512: bytes or ndarray in, uint32 ndarray (16,) out."""
+    if isinstance(x, np.ndarray):
+        x = x.tobytes()
+    return np.frombuffer(_sha3_512(x), dtype=np.uint32).copy()
```
qzhodl: The old implementation explicitly serialized/deserialized Ethash words as little-endian (`decode_int` / `encode_int` in the pre-PR version), but `np.uint32` here is native-endian. On a big-endian host this will reinterpret the Keccak output differently and change cache/dataset/mix digest generation. Can we make the byte order explicit here (for example `dtype="<u4"`), and do the same for the matching `_sha3_256` path below?

ethereum/pow/tests/test_ethash.py
```diff
@@ -117,6 +130,76 @@ def test_ethash_mining(self):
...
+        )
+        self.assertEqual(py_r[b"mix digest"], cy_r[b"mix digest"])
+        self.assertEqual(py_r[b"result"], cy_r[b"result"])
+
+    def test_rust_matches_python_fallback(self):
+        """Rust (ethash_rs) implementation matches the original hex-based baseline."""
+        try:
+            from ethereum.pow.ethash_rs import rs_calc_dataset_item, rs_hashimoto_light
+        except ImportError:
+            self.skipTest("Rust extension not built (run: python setup.py build_ext --inplace)")
```
qzhodl: These tests silently skip when the Rust extension is not built, which means the default CI path may never exercise the new `ethash_rs` fast path at all. Since this PR adds native implementations as the preferred runtime when available, can we add a CI job (or a build step in the existing workflow) that builds `ethash_rs` and runs this test with `ETHASH_LIB=ethash_rs`? Otherwise regressions here are likely to slip through until after merge.

ethereum/pow/ethash.py
```diff
@@ -1,94 +1,166 @@
...
 
-from ethereum.pow.ethash_utils import *
+from ethereum.pow.ethash_utils import (
+    ethash_sha3_512, ethash_sha3_256,
+    FNV_PRIME, HASH_BYTES, WORD_BYTES, MIX_BYTES,
+    DATASET_PARENTS, CACHE_ROUNDS, ACCESSES, EPOCH_LENGTH,
+)
+
+# uint32 overflow is intentional in FNV arithmetic
+np.seterr(over="ignore")
```
qzhodl: The overflow behavior here is intentional for FNV arithmetic, but `np.seterr(over="ignore")` changes NumPy's error handling globally at import time for the entire process, not just Ethash. Could we scope this with `np.errstate(over="ignore")` around the specific vectorized operations instead? That would avoid surprising unrelated NumPy code elsewhere in the codebase.

ethereum/pow/ethash.py
```diff
@@ -1,94 +1,166 @@
...
+                     f"Use 'ethash', 'ethash_cy', 'ethash_rs', or 'auto'.")
+
+logging.getLogger(__name__).info("ethash implementation: %s", ETHASH_LIB)
+
 
 cache_seeds = [b"\x00" * 32]  # type: List[bytes]
 
 
-def mkcache(cache_size: int, block_number) -> List[List[int]]:
+@lru_cache(2)
```
qzhodl: Non-blocking: the previous implementation cached 10 entries here, and this changes it to 2. Was that an intentional memory tradeoff?

ethereum/pow/ethash_utils.py
```diff
@@ -39,73 +26,18 @@ def _sha3_512(x):
...
-
-
-def deserialize_cache(ds):
-    return [
-        deserialize_hash(ds[i : i + HASH_BYTES]) for i in range(0, len(ds), HASH_BYTES)
-    ]
+def ethash_sha3_512(x: Union[bytes, np.ndarray]) -> np.ndarray:
+    """sha3-512: bytes or ndarray in, little-endian uint32 ndarray (16,) out."""
+    if isinstance(x, np.ndarray):
+        x = x.tobytes()
```
qzhodl: I think this only fixes half of the endianness issue. Using `dtype="<u4"` below makes the `bytes -> uint32 words` direction explicit, but when `x` is an `ndarray` we still serialize it back with native-endian `tobytes()`. Since most of the Ethash pipeline keeps intermediate state as `np.uint32` arrays, a big-endian host can still feed different byte sequences into Keccak here. The smallest fix may be to serialize ndarray inputs explicitly as little-endian too, e.g. `x = x.astype("<u4", copy=False).tobytes()`.

ethereum/pow/ethash.py
```diff
@@ -1,94 +1,166 @@
...
+        mix_r = mix.reshape(-1, 4)
+        cmix = mix_r[:, 0] * _FNV_PRIME ^ mix_r[:, 1]
+        cmix = cmix * _FNV_PRIME ^ mix_r[:, 2]
+        cmix = cmix * _FNV_PRIME ^ mix_r[:, 3]
+
+    s_cmix = np.concatenate([s, cmix])
     return {
-        b"mix digest": serialize_hash(cmix),
-        b"result": serialize_hash(ethash_sha3_256(s + cmix)),
+        b"mix digest": cmix.tobytes(),
```
qzhodl, line 161: Related to the endianness fix above: `mix digest` is an externally compared byte string, so it should probably also be serialized explicitly as little-endian instead of relying on native-endian `tobytes()`. A minimal change here would be `cmix.astype("<u4", copy=False).tobytes()`.

qzhodl (approved):

ethereum/pow/keccak_tiny.c
```diff
@@ -0,0 +1,175 @@
+/** libkeccak-tiny
+ *
+ * A single-file implementation of SHA-3 and SHAKE.
+ * Copied from https://github.com/coruus/keccak-tiny
```
blockchaindevsh: The file is copied from keccak-tiny with local modifications. The comment says "Copied from https://github.com/coruus/keccak-tiny" but doesn't record which commit or version. If a bug is found upstream, it'll be hard to determine whether this copy is affected. Consider adding the commit hash to the header comment.

ethereum/pow/ethash_cy.pyx
```diff
@@ -0,0 +1,229 @@
...
+    cdef uint32_t s_cmix[24]         # s(16) + cmix(8)
+    cdef uint32_t result_hash[8]
+    cdef Py_ssize_t cache_n = cache.shape[0]
+    cdef const uint32_t* cache_ptr = &cache[0, 0]
+
+    # nonce_rev = nonce[::-1]
+    cdef Py_ssize_t header_len = header.shape[0]
+    cdef Py_ssize_t nonce_len = nonce.shape[0]
+    cdef uint8_t seed_buf[128]  # header (up to ~80) + nonce (8)
+    memcpy(seed_buf, &header[0], header_len)
```
blockchaindevsh, line 193: Adding a guard like:

```python
if header_len + nonce_len > 128:
      raise ValueError("header + nonce too large")
```

would be more defensive.(but not a real risk given the call site is only `hashimoto_light` in ethash.py, which passes `np.frombuffer(header, dtype=np.uint8)` where header is a 32-byte block hash)

blockchaindevsh (approved):

PR #974, opened 2026-03-15
Part 1 - upgrade: update all dependencies for Python 3.13 compatibility

ping-ke:
## Summary
  - Bump all package version constraints to versions that support Python 3.13
  - Replace unmaintained/incompatible packages:
    - `python-rocksdb` → `rocksdict`
    - `jsonrpcserver` + `jsonrpcclient` → `httpx` + custom client
    - `upnpclient` + `netifaces` → `async-upnp-client`
    - `pyethash` (C extension, dropped)
    - `jsonrpc-async` → `jsonrpc_async`
  - Set `python_requires = ">=3.13"` in setup.py

  ## Test plan
  - [x] `pip install -r requirements.txt` succeeds on Python 3.13
  - [ ] All other upgrade/* PRs apply cleanly on top of this


requirements.txt
```diff
@@ -1,34 +1,31 @@
...
+aioprocessing>=2.0
+cachetools>=5
+coincurve>=19.0.1
+decorator>=5
+ecdsa>=0.18
+eth-hash[pycryptodome]>=0.5,<0.6
+eth-utils>=2.3
+eth-keys>=0.5
+eth-bloom>=2.0
+httpx>=0.28
```
qzhodl, line 12: It added httpx as a new dependency to replace jsonrpcserver/jsonrpcclient. However, aiohttp is already a project dependency — used for the web server in quarkchain/cluster/jsonrpc.py and as an HTTP client in several tool files (fund_testnet.py, batch_deploy_contract.py, monitoring.py).

aiohttp provides HTTP client functionality that could serve the same role as httpx for building the custom JSON-RPC client. Adding httpx introduces a second HTTP library into the dependency tree.

That said, httpx does have some advantages: cleaner requests-like API, built-in sync+async support (useful if any tool scripts need synchronous calls), and HTTP/2 support. If there's a specific reason httpx was chosen over aiohttp, it would be good to document that in the PR description. Otherwise, using aiohttp for the client side as well would keep the dependency list smaller.

requirements.txt
```diff
@@ -1,34 +1,31 @@
...
-jsonrpc-async==0.6
-numpy==1.15.1
-psutil==5.6.6
-python-rocksdb==0.6.9
-requests==2.20.0
-aiohttp_cors==0.7.0
-eth-utils==1.2.0
-eth-keys==0.2.0b3
-eth-bloom==1.0.0
-pyethash>=0.1.27,<1.0.0
```
qzhodl, line 18: Dropping pyethash is reasonable since it's a C extension last released in 2015 with no Python 3.13 support. The pure-Python fallback in ethereum/pow/ethpow.py (lines 14-16) will keep all functionality working.

For production nodes, this is a non-issue: production uses REMOTE_MINE = True with an external miner, so the node only verifies PoW results — one hashimoto call per block, which the pure-Python implementation handles fine.

However, the default config has REMOTE_MINE = False (quarkchain/config.py line 110). In this mode, the local mining loop (miner.py line 343) calls hashimoto thousands/millions of times per block. With the pure-Python fallback, dev/testing scenarios using local mining will be significantly slower than before.

Suggestions:

 - Consider adding a more visible warning (e.g., log at WARNING level instead of ImportWarning) when falling back to pure-Python ethash, so developers are aware of the performance impact
 - Document in the PR that local mining performance is degraded without pyethash
 - If a Python 3.13-compatible C ethash library becomes available in the future, it can be re-added as an optional dependency

requirements.txt
ping-ke: httpx was chosen because it provides both sync and async clients with a unified API, keeping jsonrpc_client.py simple and consistent. aiohttp is async-only so the sync client would need urllib.request — meaning two different HTTP implementations in one file. httpx as a pure client library is lightweight (~200KB with deps) and doesn't overlap with aiohttp's server role.

requirements.txt
```diff
@@ -10,12 +10,12 @@ eth-utils>=2.3
 eth-keys>=0.5
 eth-bloom>=2.0
 httpx>=0.28
-jsonrpc_async>=2.1.2
 numpy>=1.26
 psutil>=6
 rocksdict
 requests>=2.28.0
 py_ecc>=7
+pyethash>=0.1.27,<1.0.0
```
qzhodl: why adding pyethash back?

requirements.txt
ping-ke: Add issue https://github.com/QuarkChain/pyquarkchain/issues/976 to check this and also add comment to PR https://github.com/QuarkChain/pyquarkchain/pull/973.

requirements.txt
ping-ke: The related comment has been added to quarkchain\jsonrpc_client.py in the PR https://github.com/QuarkChain/pyquarkchain/pull/971.

qzhodl (approved):

requirements.txt
```diff
@@ -1,34 +1,30 @@
...
+decorator>=5
+ecdsa>=0.18
+eth-hash[pycryptodome]>=0.5,<0.6
+eth-utils>=2.3
+eth-keys>=0.5
+eth-bloom>=2.0
+httpx>=0.28
+numpy>=1.26
+psutil>=6
+rocksdict
```
syntrust: Why is there no version scope for rocksdict?

ping-ke: added version.

syntrust:
It seems all commit-build-test checks failed due to
>ModuleNotFoundError: No module named 'rocksdb'

Is it possible that it is still being used somewhere?

ping-ke:
> It seems all commit-build-test checks failed due to
> 
> > ModuleNotFoundError: No module named 'rocksdb'
> 
> Is it possible that it is still being used somewhere?

Because the entire modification is very large, the individual PRs cannot pass the tests on their own; they need to be merged to pass the tests.

syntrust (approved):

requirements.txt
```diff
@@ -1,34 +1,33 @@
...
-eth-keys==0.2.0b3
-eth-bloom==1.0.0
-pyethash>=0.1.27,<1.0.0
-py_ecc==1.4.3
-eth-hash[pycryptodome]==0.1.4
-pytest-timeout==1.3.3
-cachetools==3.1.1
-websockets==8.0.2
+aiohttp>=3.9
+aiohttp_cors>=0.7.0
```
blockchaindevsh: All version constraints are open-ended. The old file pinned exact versions (aiohttp==3.1.3). Open-ended constraints risk breakage from future major version bumps. Consider using compatible release operators (~=) or upper-bounded ranges for critical deps, e.g., aiohttp>=3.9,<4 or cryptography>=42,<44.

mainnet/singularity/Dockerfile
```diff
@@ -1,9 +1,40 @@
...
+    libzstd-dev \
+    zlib1g-dev \
     vim \
   && rm -rf /var/lib/apt/lists/*
+
+# Python
+RUN pip install --upgrade pip
+
+# Rust
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
```
blockchaindevsh, line 29: This adds ~500MB+ to the image for the Rust toolchain, which is only needed at build time. Consider a multi-stage build: build extensions in a builder stage, then copy the .so files to a slim runtime image.

requirements.txt
```diff
@@ -1,34 +1,33 @@
...
+ecdsa>=0.18
+eth-hash[pycryptodome]>=0.5,<0.6
+eth-utils>=2.3
+eth-keys>=0.5
+eth-bloom>=2.0
+httpx>=0.28
+Cython>=3
+maturin>=1.4
+setuptools-rust>=1.9
+setuptools>=80.8.0
```
blockchaindevsh:
```
Cython>=3
 maturin>=1.4
 setuptools-rust>=1.9
 setuptools>=80.8.0
```

These are build-time dependencies, not runtime. They should be in pyproject.toml under [build-system] requires or a separate requirements-dev.txt.

ping-ke: Acknowledged. However, since the project uses setup.py (not pyproject.toml), moving them to [build-system] requires wouldn't take effect for our build flow (python setup.py build_ext --inplace). A requirements-dev.txt split is possible, but given the project's scale we prefer keeping everything in one file for simplicity.
  Additionally, with the Dockerfile now downloading pre-built .so files from S3, these build-time packages are no longer installed in the runtime image at all.

mainnet/singularity/Dockerfile
ping-ke: Adding a tiny version of Dockerfile which removed all build environments, it significantly reduce the docker  image size, but Docker with build environments is more convenient to use, so I kept the current version.

blockchaindevsh:
image

Looks like the merge order matters, is this intentional?

mainnet/singularity/Dockerfile-tiny
```diff
@@ -0,0 +1,35 @@
+FROM python:3.13-slim
```
blockchaindevsh, line 1: Maybe a build arg in the original Dockerfile is cleaner than maintaining two separate files ?

ping-ke:
> image
> Looks like the merge order matters, is this intentional?

We will merge all the PRs to [upgrade/py313-baseline](https://github.com/QuarkChain/pyquarkchain/tree/upgrade/py313-baseline) branch. then run unittest and run node to test it before merge to master.

mainnet/singularity/Dockerfile
```diff
@@ -1,25 +1,71 @@
...
+  && rustup toolchain install stable \
+  && rustup default stable \
+  && python setup.py build_ext --inplace
+
+# ---- prebuilt-true: download pre-built extensions from S3 ----
+FROM base AS prebuilt-true
+
+RUN curl -fsSL https://s3-us-west-2.amazonaws.com/pyqkcmainnet/ethash_cy.cpython-313-x86_64-linux-gnu.so \
+         -o /code/pyquarkchain/ethereum/pow/ethash_cy.cpython-313-x86_64-linux-gnu.so \
+  && curl -fsSL https://s3-us-west-2.amazonaws.com/pyqkcmainnet/ethash_rs.cpython-313-x86_64-linux-gnu.so \
```
blockchaindevsh, line 73: Is it ok to support only x86_64 (Is ARM supposed to be supported)?

blockchaindevsh (approved):
The commit message("support arm64 builds") is a bit misleading.

A more accurate message would be something like:

Dockerfile: fail explicitly when PREBUILT=true on non-amd64

PR #979, opened 2026-05-14
Add pyethash C extension backend with zero-copy cache and cross-impl tests

ping-ke:
### Summary

- Upgrade `pyethash` to support Python 3.13 (via `QuarkChain/ethash` `upgrade-py3-13` branch)
- Add `pyethash` as a selectable ethash backend via `ETHASH_LIB=pyethash`
- Halve per-epoch memory usage by making the cache array a zero-copy view
- Fix Docker base image to include `gcc` required to build the C extension
- Add regression tests covering epoch 520 (the epoch that caused a sync failure) and a random epoch
- Extend `bench_hashimoto_compare` to include pyethash in all comparisons

### Changes

**`ethereum/pow/ethash.py`**
Use `np.frombuffer` without `.copy()` so the cache array is a read-only view into
the raw `bytes` returned by `pyethash.mkcache_bytes`. Both share the same ~16 MB buffer
instead of allocating ~32 MB per epoch. Safe because `calc_dataset_item` always copies
a row before mutating it. Also adds comments explaining why `hashimoto_light` calls
`_get_pyethash_cache` a second time (O(1) LRU hit, not a recomputation).

**`mainnet/singularity/Dockerfile`**
Move `gcc` to the base stage so it is available when `pip install pyethash` compiles
the C extension. Remove redundant `python3-dev` from the build stage.

**`requirements.txt`**
Pin pyethash to the `upgrade-py3-13` branch of `QuarkChain/ethash`. Update comment
to say "requires gcc" instead of "requires C compiler". `upgrade-py3-13` branch will be removed
after merge to master.

**`ethereum/pow/ethash.py` / `ethash_cy.pyx` / `ethash_rs/`**
Remove `ethash_cy` (Cython) and `ethash_rs` (Rust) backends along with all associated build files and bindings. `pyethash` becomes the sole C-extension backend.

**`ethereum/pow/tests/test_ethash.py`**
Remove `ethash_cy` and `ethash_rs` and their associated tests. Add `TestEthashPyethashEpochs`:
builds caches and runs `hashimoto_light` with `pyethash` at epoch 520 and a random epoch,
asserting correct `mix digest` and `result`.
Pure-Python (`ethash`) is excluded — correctness is already covered by the existing
`test_pyethash_matches_python_fallback` tests; its performance at large epochs
(epoch-520 cache ≈ 84 MB) makes inclusion impractical.

#### Python 3.8 vs Python 3.13 — pyquarkchain Comparison

> pyethash (C extension) performance is essentially unchanged between Python versions.
> Pure Python is dramatically faster in 3.13 due to interpreter improvements, but pyethash
> already dominates it in both versions.

**Micro-benchmark: `bench_hashimoto_compare` (epochs 0–29, 100 calls/epoch)**

*pyethash — Python 3.8 vs 3.13 (C extension, performance nearly identical):*

| Version | mkcache (ms) | hashimoto (ms) | check_pow (ms) | 
|---|---|---|---|
| Python 3.8 | 356 | 0.8 | 4.5 |
| Python 3.13 | 363 | 0.8 | 0.8 | 
| speedup (3.13 vs 3.8) | ~1× | ~1× | **~6×** |

`mkcache` and `hashimoto` are C-bound and unchanged. `check_pow` is ~6× faster in 3.13 due to reduced Python wrapper overhead around the C call.

*Pure Python — Python 3.8 vs 3.13 (interpreter improvements):*

| Version | mkcache (ms) | hashimoto (ms) | 
|---|---|---|
| Python 3.8 | 64,192 | 120.6 |
| Python 3.13 | 7,571 | 44.7 | 
| speedup (3.13 vs 3.8) | **~8.5×** | **~2.7×** |

*Python 3.13 — pyethash vs pure Python:*

| Implementation | mkcache (ms) | hashimoto (ms) | check_pow (ms) |
|---|---|---|---|
| pyethash | 363 | 0.8 | 0.8 |
| pure Python | 7,571 | 44.7 | 44.9 |
| speedup | **21×** | **57×** | **60×** |

Even with Python 3.13's interpreter gains, pyethash remains decisively faster.

**Node sync — large root block sync time (s/block)**

| Block | Python 3.13 | Python 3.8 |
|---|---|---|
| 230,386 | 0.28 | 0.34 |
| 230,434 | 0.30 | 0.34 |
| 230,435 | 0.31 | 0.35 |
| 230,458 | 0.29 | 0.33 |
| 230,459 | 0.28 | 0.33 |
| 230,471 | 0.30 | 0.34 |
| 230,530 | 0.37 | 0.34 |
| 230,531 | 0.36 | 0.34 |
| 230,579 | 0.29 | 0.35 |
| 230,590 | 0.30 | 0.34 |
| 230,597 | 0.31 | 0.32 |
| 230,598 | 0.29 | 0.33 |
| **avg** | **0.31** | **0.34** |

Python 3.13 is ~9% faster per large root block on average.

**Node sync — full range (blocks 20,000 – 220,000)**

| Version | Total time |
|---|---|
| Python 3.8 | 8 h 33 m 51 s |
| Python 3.13 | 8 h 15 m 18 s |

Python 3.13 finishes the full range ~18 minutes (~3.7%) faster than Python 3.8, consistent with
the per-block large-root advantage seen above.

### Test plan

- [x] `python -m pytest`
- [x] `python -m ethereum.pow.tests.bench_hashimoto_compare`
- [x] Docker build: `docker buildx build --build-arg GIT_TAG=update/pyethash -t qkc:pyethash .`
- [ ] Node synchronization (on-going)


ethereum/pow/ethash.py
```diff
@@ -79,17 +113,36 @@ def _get_cache(seed: bytes, n: int) -> np.ndarray:
...
         cache_seeds.append(new_seed)
 
-    seed = cache_seeds[block_number // EPOCH_LENGTH]
+    seed = cache_seeds[n]
     return _get_cache(seed, cache_size // HASH_BYTES)
 
 
 def hashimoto_light(
-    full_size: int, cache: np.ndarray, header: bytes, nonce: bytes
+    full_size: int, cache: np.ndarray, header: bytes, nonce: bytes, block_number: int = 0,
```
blockchaindevsh: Why not always pass `block_number` explicitly, will the default value cause footgun?

qzhodl:
- In the end, we may want to keep only the pure Python implementation and the pyethash implementation to simplify the overall change.
 - For benchmarking, we may want to compare the Python 3.8 version with the current version.

requirements.txt
```diff
@@ -31,3 +28,6 @@ rlp>=3,<5
 
 # pyethapp/accounts.py dependency
 pbkdf2>=1.3,<2
+
+# pyethash C extension for PoW (set ETHASH_LIB=pyethash to activate; requires gcc)
```
blockchaindevsh: Looks like this comment is stale, there's no `ETHASH_LIB` environment variable now?

ethereum/pow/ethpow.py
```diff
@@ -29,6 +29,7 @@ def check_pow(
 
     if is_test:
         cache_size, full_size = 1024, 32 * 1024
+        ethash.set_ethash_lib("python")  # force python impl for non-canonical size
```
blockchaindevsh: Is this needed when `mkcache_pyethash` and `hashimoto_light_pyethash` already fall back to python when sizes are non-canonical?

ethereum/pow/ethash_utils.py
```diff
@@ -50,6 +51,7 @@ def isprime(x):
     return True
 
 
+@lru_cache(maxsize=256)
```
blockchaindevsh, line 54: LRU keyed by block_number, not epoch — wastes cache slots for blocks in the same epoch. Cheap fix: cache by `block_number // EPOCH_LENGTH`.

ethereum/pow/ethash_utils.py
```diff
@@ -58,6 +60,7 @@ def get_cache_size(block_number):
     return sz
 
 
+@lru_cache(maxsize=256)
```
blockchaindevsh, line 63: LRU keyed by block_number, not epoch — wastes cache slots for blocks in the same epoch. Cheap fix: cache by `block_number // EPOCH_LENGTH`.

blockchaindevsh:
Looks like the PR description and README.md(Mentions ethash_rs, ethash_cy, and ETHASH_LIB override. None of these exist now) are outdated?

ethereum/pow/ethpow.py
ping-ke: Only `mkcache_pyethash` and `hashimoto_light_pyethash` support is_test (1024 and 32 * 1024 size). So when is_test, we need to `set_ethash_lib` to python.

ping-ke: Only `mkcache_pyethash` and `hashimoto_light_pyethash` support is_test (1024 and 32 * 1024 size). So when is_test, we need to `set_ethash_lib` to python.

ping-ke:
> Looks like the PR description and README.md(Mentions ethash_rs, ethash_cy, and ETHASH_LIB override. None of these exist now) are outdated?

Updated.

ethereum/pow/ethpow.py
```diff
@@ -81,15 +83,17 @@ def mine(
     rounds: int = 1000,
     is_test: bool = False,
 ) -> Tuple[Optional[bytes], Optional[bytes]]:
+    epoch = block_number // EPOCH_LENGTH
     if is_test:
         cache_size, full_size = 1024, 32 * 1024
+        ethash.set_ethash_lib("python")  # force python impl for non-canonical size
```
qzhodl: This switches the ethash implementation globally when `is_test=True`, but never switches it back. After one test-mode PoW call, the same process will keep using the pure Python backend for normal PoW as well. Since the pyethash wrappers already fall back for non-canonical sizes, can we avoid mutating global backend state here?

requirements.txt
```diff
@@ -31,3 +28,5 @@ rlp>=3,<5
 
 # pyethapp/accounts.py dependency
 pbkdf2>=1.3,<2
+
+pyethash @ git+https://github.com/QuarkChain/ethash.git
```
qzhodl: This VCS dependency is still floating on the repository default branch. For reproducible installs/builds, can we pin it to a specific commit SHA or tag?

quarkchain/cluster/master.py
```diff
@@ -718,9 +718,9 @@ async def handle_add_minor_block_header_list_request(self, req):
             self.master_server.root_state.add_validated_minor_block_hash(
                 minor_block_header.get_hash(), coinbase_amount_map.balance_map
             )
-            Logger.info(
-                "adding {} mblock to db".format(minor_block_header.get_hash().hex())
-            )
+#            Logger.info(
```
qzhodl: This log change seems unrelated to the pyethash backend work. Can we either split it out or add the rationale to the PR description?

qzhodl (approved):

syntrust (approved):

ethereum/pow/tests/test_ethash.py
```diff
@@ -206,3 +141,86 @@ def test_pyethash(self):
...
+
+        # Compare hashimoto_light output: numpy Python path vs pyethash C path
+        py_r = hashimoto(header, nonce, full_size, lambda x: calc_dataset_item(cache, x))
+        pe_r = _pyethash.hashimoto_light(0, raw, header, 0)
+
+        self.assertEqual(py_r[b"mix digest"], pe_r[b"mix digest"])
+        self.assertEqual(py_r[b"result"],     pe_r[b"result"])
+
+
+class TestEthashCrossImplEpochs(unittest.TestCase):
```
blockchaindevsh, line 181: class name is misleading — no cross-impl comparison happens, only pyethash. Rename.

PR #981, opened 2026-05-22
tools: add scan.py for on-chain TX and active-user analytics

ping-ke:
## Summary

Adds `quarkchain/tools/scan.py`, a standalone DB scanner that reads shard RocksDB databases directly (no running node required) and outputs transaction count and active-user metrics for a given time range in Markdown format.

## Features

- **Time-range input** — accepts `--start` / `--end` as UTC datetime strings
- **Auto-discovers shards** — globs `shard-*.db` under `--db-path`, handles any number of shards automatically
- **Efficient block lookup** — exponential + binary search on `mi_<height>`
  index keys to locate the block-height range in O(log N), avoiding a full DB scan
- **Four aggregation views**
  | View | Description |
  |------|-------------|
  | Daily | Per-calendar-day TX count and unique active users |
  | Monthly | Per-calendar-month totals |
  | 3-Month (Quarter) | Calendar quarters Q1–Q4 |
  | Rolling 3-Month | Sliding window: each month M shows M-2 + M-1 + M combined |
- **Active users = senders only** — counts the 20-byte sender address per transaction; avoids inflating numbers with contract addresses or cross-shard recipients
- **Markdown output** — prints to stdout by default; `--output <file>` writes to disk

## Usage

```bash
python quarkchain/tools/scan.py \
  --db-path /data/qkc \
  --start "2026-01-01 00:00:00" \
  --end   "2026-05-01 00:00:00" \
  --output report.md

Notes

- Rolling 3-Month windows that extend before --start will show 0 for the
out-of-range months; widen --start by 2 months to fill those windows.

### DIFF
--- a/quarkchain/tools/scan.py
+++ b/quarkchain/tools/scan.py
@@ -1,0 +1,309 @@
+#!/usr/bin/env python3
+"""
+scan.py - QuarkChain DB Scanner
+
+Scans shard RocksDB databases for a time range and outputs TX count and
+active user metrics in Markdown format.
+
+Usage:
+    python scan.py --db-path /data/qkc --start "2026-01-01 00:00:00" --end "2026-05-01 00:00:00"
+"""
+
+import argparse
+import glob
+import os
+import sys
+from collections import defaultdict
+from datetime import datetime, timezone
+
+# Allow running from any directory
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../.."))
+
+from quarkchain.db import PersistentDb
+from quarkchain.core import MinorBlock
+
+
+# ---------------------------------------------------------------------------
+# DB helpers
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by syntrust** [comment]
As the PR for the Python upgrade is not merged, it is preferable to specify the Python version required to run this tool.

File: quarkchain/tools/scan.py [hunk 1]
```diff
@@ -0,0 +1,334 @@
...
+
+import argparse
+import glob
+import os
+import sys
+from collections import defaultdict
+from datetime import datetime, timezone
+
+# Allow running from any directory
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../.."))
```
qzhodl: ../../.. goes one level above the repo root, so import quarkchain fails with ModuleNotFoundError when running python quarkchain/tools/scan.py directly; should be ../..

qzhodl (approved):

syntrust (approved):

PR #984, opened 2026-06-09
Fix async task reference leaks and unclosed HTTP clients

ping-ke:
### Problem

  Two categories of resource management issues found during code review of the py313 upgrade:

  1. asyncio.create_task() / ensure_future() return values discarded — the event loop holds only weak references to tasks. An unsaved task can be garbage-collected mid-execution with no error or log output, causing silent failures:
    - Synchronizer.__run() (master + shard): if GC'd, self.running stays True permanently → node silently stops syncing, no restart, no error
    - miner: recursive _mine_new_block_async() call inside handle_mined_block discards the returned task → mining silently stops after a block is found
    - tx_generator: generator task GC'd mid-loop → self.running stuck True, loadtest permanently blocked
    - p2p_manager: server loop task has no saved reference
  2. AsyncJsonRpcClient (wrapping httpx.AsyncClient) not closed at call sites — leaks connection pool resources; long-running processes like monitoring.py will eventually exhaust file descriptors

###  Changes

  quarkchain/jsonrpc_client.py
  - Add __aenter__ / __aexit__ to AsyncJsonRpcClient for async with support

  quarkchain/p2p/p2p_manager.py
  - Save create_task(self.server.run()) to self._server_task

  quarkchain/cluster/master.py
  - Synchronizer.add_task: save ensure_future(self.__run()) to self._run_task
  - _main_async: save check_db task to local variable; await start_mining() directly instead of create_task

  quarkchain/cluster/shard.py
  - Synchronizer.add_task: same fix as master — save ensure_future(self.__run()) to self._run_task

  quarkchain/cluster/miner.py
  - Anchor recursive _mine_new_block_async() tasks via self._pending_block_tasks set with add_done_callback(discard); self._mining_task is intentionally unchanged
  (used by disable() for cancellation)

  quarkchain/cluster/tx_generator.py
  - Save create_task(self.__gen(...)) to self._gen_task

  quarkchain/cluster/tests/test_jsonrpc.py
  - send_request: use async with AsyncJsonRpcClient(...)

  quarkchain/tools/monitoring.py
  - async_watch: add try/finally to close long-lived clients on loop exit
  - fetch_peers_async: move server.close() into finally — CancelledError is a BaseException and bypasses except Exception

  quarkchain/tools/adjust_difficulty.py, fund_testnet.py, batch_deploy_contract.py
  - Add try/finally or async with to close AsyncJsonRpcClient after use


quarkchain/jsonrpc_client.py
```diff
@@ -64,4 +64,10 @@ async def call(self, method, *params):
...
     async def close(self):
-        await self.client.aclose()        
\ No newline at end of file
+        await self.client.aclose()
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *exc):
+        await self.client.aclose()
```
blockchaindevsh: I'd suggest:

```python
  async def __aexit__(self, exc_type, exc, tb):
      await self.close()
```

so if one day we want to change the close behavior, we only need to change one place instead of two.

quarkchain/tools/adjust_difficulty.py
```diff
@@ -101,25 +105,33 @@ async def adjust_imbalanced_hashpower(args):
...
+        try:
+            await asyncio.gather(
+                *[
+                    async_adjust(idx, server, rich_root, rich_minor, not args.do_not_mine)
+                    for (idx, server) in servers_rich
+                ]
+            )
+        finally:
+            for _, server in servers_rich:
+                await server.close()
```
blockchaindevsh: partial cleanup on rich-gather failure.

`servers_poor` is allocated before the rich-cluster gather. If the rich gather raises, only `servers_rich` gets cleaned up by its try/finally; the `len(servers_poor)` clients leak. Since `adjust_imbalanced_hashpower` is a one-shot CLI script the process exits soon anyway, but it's still a regression vs. the stated PR intent.
                                                                                                                                                                                                                                                    
  Fix: use `contextlib.AsyncExitStack` to register both server lists at allocation, or move servers_poor = [...] after the rich gather succeeds.

blockchaindevsh:
One more place:

image

image

ping-ke:
> One more place:
> 
> image image

Updated.

blockchaindevsh (approved):

quarkchain/cluster/simple_network.py
```diff
@@ -472,9 +473,11 @@ async def connect_seed(self, ip, port):
 
         Logger.info("connecting {} peers ...".format(len(resp.peer_info_list)))
         for peer_info in resp.peer_info_list:
-            asyncio.create_task(
+            t = asyncio.create_task(
                 self.connect(str(ipaddress.ip_address(peer_info.ip)), peer_info.port)
             )
+            self._connect_tasks.add(t)
```
syntrust, line 489: There is no cancel `_connect_tasks` in shutdown(), so these connection tasks may continue to run after shutdown.

qzhodl:
We may want to separate this PR into two since it is resolving two issues together

quarkchain/cluster/subscription.py
```diff
@@ -85,9 +85,11 @@ async def notify_sync(self, data: Optional[Tuple[int, ...]] = None):
                 "currentBlock": tip_height,
                 "highestBlock": highest_block,
             }
+        tasks = []
         for sub_id, websocket in self.subscribers[SUB_SYNC].items():
             response = self.response_encoder(sub_id, result)
-            asyncio.create_task(websocket.send(json.dumps(response)))
+            tasks.append(websocket.send(json.dumps(response)))
+        await asyncio.gather(*tasks)
```
qzhodl: The function used to create a task and return immediately, but now it awaits the task instead. This changes the behavior. Is that intentional?

qzhodl (approved):

quarkchain/cluster/subscription.py
ping-ke: Changed back.

PR #987, opened 2026-06-10
fix(P3-8): skip dead connections when adding peers after active_event

ping-ke:
### Summary
`active_and_loop_forever`'s `finally` block now sets `active_event` even on an early close, so
callers awaiting `active_event.wait()` can no longer assume the connection is live. Both
peer-registration call sites are guarded with `conn.is_active()` before `add_peer()`.

### Related Issue

### Changes
- **`shard.py` `__init_connections`:** guard peer-shard connections received from master with
  `if conn.is_active()` before `add_peer`.
- **`slave.py` `handle_create_cluster_peer_connection`:** same guard for cluster peer connections,
  with a comment noting that `active_event` is also set on early close.

### Files
`quarkchain/cluster/shard.py`, `quarkchain/cluster/slave.py`

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -597,7 +597,8 @@ class Shard:
             conns.append(peer_shard_conn)
         await asyncio.gather(*[conn.active_event.wait() for conn in conns])
         for conn in conns:
-            self.add_peer(conn)
+            if conn.is_active():
+                self.add_peer(conn)
 
     async def __init_genesis_state(self, root_block: RootBlock):
         block, coinbase_amount_map = self.state.init_genesis_state(root_block)
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -599,6 +599,13 @@ class Shard:
         for conn in conns:
             if conn.is_active():
                 self.add_peer(conn)
+            else:
+                Logger.warning(
+                    "cluster peer {} vconn for shard {} closed before becoming "
+                    "active, skipped".format(
+                        conn.cluster_peer_id, self.full_shard_id
+                    )
+                )
 
     async def __init_genesis_state(self, root_block: RootBlock):
         block, coinbase_amount_map = self.state.init_genesis_state(root_block)
--- a/quarkchain/cluster/simple_network.py
+++ b/quarkchain/cluster/simple_network.py
@@ -131,6 +131,16 @@ class Peer(P2PConnection):
         self._loop_task = asyncio.create_task(self.active_and_loop_forever())
         await self.wait_until_active()
 
+        # wait_until_active() also returns when the connection closed before ever
+        # becoming active (active_event is set in the finally block of
+        # active_and_loop_forever). Adding a CLOSED peer to the pools would leak it:
+        # close() only removes ACTIVE peers, so broadcasts/get_peer_list would keep
+        # using it forever. close_dead_peer() removes it from the pools (if present)
+        # and tears down the virtual shard connections created above.
+        if not self.is_active():
+            self.close_dead_peer()
+            return "peer {} closed before becoming active".format(self.id.hex())
+
         # Only make the peer connection avaialbe after exchanging HELLO and creating virtual shard connections
         self.network.active_peer_pool[self.id] = self
         self.network.cluster_peer_pool[self.cluster_peer_id] = self
--- a/quarkchain/cluster/slave.py
+++ b/quarkchain/cluster/slave.py
@@ -353,9 +353,12 @@ class MasterConnection(ClusterConnection):
         # wait for all the connections to become active before return
         await asyncio.gather(*active_futures)
 
-        # Make peer connection available to shard once they are active
+        # Make peer connection available to shard once they are active.
+        # Skip connections that died before becoming active (active_event is also
+        # set in the finally block of active_and_loop_forever on early close).
         for shard, peer_shard_conn in shard_to_conn.items():
-            shard.add_peer(peer_shard_conn)
+            if peer_shard_conn.is_active():
+                shard.add_peer(peer_shard_conn)
 
         return CreateClusterPeerConnectionResponse(error_code=0)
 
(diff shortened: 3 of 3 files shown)


qzhodl:
There is a third call site with the same issue that this PR misses, and it is
actually the worst one: `Peer.start` in `simple_network.py` (line ~131).

After `await self.wait_until_active()`, the peer is unconditionally added to
`active_peer_pool` and `cluster_peer_pool`. If the connection died before
becoming active, a CLOSED peer enters both pools. Worse, `Peer.close()` only
removes the peer from the pools when `state == ACTIVE`, so this dead peer is
never cleaned up: broadcasts keep iterating over it, `get_peer_list` keeps
advertising it, and `destroy_peer_cluster_connections` never runs.

quarkchain/cluster/shard.py
```diff
@@ -597,7 +597,8 @@ async def create_peer_shard_connections(self, cluster_peer_ids, master_conn):
             conns.append(peer_shard_conn)
         await asyncio.gather(*[conn.active_event.wait() for conn in conns])
         for conn in conns:
-            self.add_peer(conn)
+            if conn.is_active():
```
qzhodl, line 600: When a dead connection is skipped here, it happens silently and the response
still returns `error_code=0`, so the master believes all virtual connections
for this peer were created. Debugging "why does this peer have no connection
on shard X" later will be hard.

Suggest at least logging it


```suggestion
    if peer_shard_conn.is_active():
        shard.add_peer(peer_shard_conn)
    else:
        Logger.warning(
            "cluster peer {} vconn for shard {} closed before becoming active, skipped".format(
                req.cluster_peer_id, shard.full_shard_id
            )
        )
```

qzhodl (approved):

blockchaindevsh (commented):
From Fable:

  Review: PR #987 — fix(P3-8): skip dead connections when adding peers after active_event

  Metadata: ping-ke → upgrade/py313-baseline (head 1b6053ed), 2 commits, 3 files, +31/−3. CI: all 4 checks green.

  Overview

  On the py3.13 baseline branch, AbstractConnection.active_and_loop_forever's finally block (quarkchain/protocol.py:204-207, already in the PR base) sets active_event even when a connection dies before ever becoming ACTIVE — so waiters no
  longer hang, but await active_event.wait() also no longer implies the connection is live. This PR fixes the three call sites that still assumed liveness:

  1. simple_network.py:134-142 Peer.start — after wait_until_active(), if the peer isn't ACTIVE, call close_dead_peer() and return an error instead of inserting the peer into active_peer_pool/cluster_peer_pool.
  2. shard.py:599-608 create_peer_shard_connections — guard add_peer() with conn.is_active(), log a warning otherwise.
  3. slave.py:355-369 handle_create_cluster_peer_connection_request — same guard, same warning.

  Correctness — verified, the fix is sound

  - The premise is real. protocol.py:204-211 sets the event in finally and forces state to CLOSED before the waiter resumes, so after the wait, not is_active() ⇔ CLOSED — the guard's dichotomy is exact.
  - The leak being fixed is real. Peer.close() (simple_network.py:155) only removes a peer from the pools when state == ACTIVE. Pre-fix, a peer that died between HELLO and activation was inserted into the pools already CLOSED with its loop task
  finished — nothing would ever remove it, and it would be served forever via handle_get_peer_list_request (simple_network.py:191-198). A genuine permanent leak plus peer-discovery pollution, triggerable by a remote peer closing right after
  HELLO.
  - No TOCTOU races. None of the three sites has an await between the is_active() check and the pool insertion, so the guard can't go stale under asyncio's single-threaded model.
  - close_dead_peer() path is safe. Its assert self.id is not None and the self.id.hex() in the new error string are fine because self.id/self.ip/self.port are assigned at simple_network.py:102-105, before the new guard. Both Peer.start callers
  handle the error return correctly.
  - Skipped vconns don't leak elsewhere. VirtualConnection doesn't register itself on the proxy connection; shard.peers via add_peer is the only registry, so a skipped dead vconn is garbage-collected. Later master-forwarded traffic hits the
  existing NULL_CONNECTION fallback (slave.py:130-146) — a tolerated-by-design path.
  - Minor harmless redundancy: if the connection went ACTIVE then closed before Peer.start resumed, destroy_peer_cluster_connections runs twice (double DESTROY broadcast). The slave handler is idempotent, so it's a wasted command at worst.
  - Remaining unguarded waiters — acceptable. master.py:880, master.py:912, and slave.py:881 still call wait_until_active() without an is_active() check, but those are trusted intra-cluster connections that fail fast right after
  (write_rpc_request raises on a non-ACTIVE connection). Fine to leave; follow-up note at most.

  Code quality / conventions

  - The new explanatory comments are accurate against the actual code (each claim was verified).
  - The PR description is stale: its "Files" section lists only shard.py and slave.py, but the most substantive change — the Peer.start dead-peer cleanup added in the "resolve comments" commit — is in simple_network.py.
  - Nit: the new warnings log the raw full_shard_id int, while neighboring code logs Branch(self.full_shard_id).to_str().
  - Reviewer qzhodl's concern (silent skip + error_code=0 makes the master believe all vconns exist) was addressed with warnings rather than an error code. Given the NULL_CONNECTION fallback, returning success is acceptable, though a
  partial-failure field in CreateClusterPeerConnectionResponse would be the stronger fix if this recurs.

  Test coverage

  No tests added. CI is green but only exercises happy paths — nothing simulates a connection dying between HELLO and activation, so all three new branches are untested. A focused unit test (close the transport before the loop task is scheduled
  in Peer.start, then assert the pools are empty and a DESTROY was broadcast) would be cheap and would lock in the contract created by the protocol.py finally change.

  Performance / security

  O(1) guards, no hot-path impact. Net security positive: this closes a vector where a malicious peer could pollute active_peer_pool with permanently-dead entries served to other nodes via GET_PEER_LIST.

  Verdict: Approve with minor changes

  1. Update the PR description to include quarkchain/cluster/simple_network.py (it currently misrepresents the change).
  2. Add a regression test for the Peer.start dead-peer path (strongly recommended, can be a fast follow-up).
  3. Optional: track the unguarded wait_until_active() sites (master.py:880/912, slave.py:881) as a follow-up, and use Branch(...).to_str() in the new log lines.

blockchaindevsh (approved):

PR #988, opened 2026-06-11
fix: JSON-RPC protocol compliance + protocol tests

ping-ke:
### Summary
Brings the HTTP and WebSocket JSON-RPC servers into spec compliance for malformed input,
notifications, and batch requests, and adds a protocol test suite. Also hardens the JSON-RPC client
to validate response ids.

### Related Issue 

### Changes
- **`jsonrpc.py` (HTTP/WS `__handle`):** return `-32700 Parse error` on malformed JSON instead of
  swallowing the error and dispatching an empty dict. Drop the dead `d = dict()` fallback and guard
  non-dict payloads with `isinstance`.
- **`jsonrpc_server.py` (`dispatch`):** suppress responses for notifications (requests with no
  `"id"`), even on the error paths. Batch (JSON array) requests are intentionally still unsupported
  but now rejected cleanly with `-32600` instead of crashing with an `AttributeError` / HTTP 500.
- **`jsonrpc_client.py`:** `JsonRpcClient` / `AsyncJsonRpcClient` now raise `JsonRpcError(-32600)`
  when the response `id` doesn't match the request `id`; also fixes a missing trailing newline.
- **`tests/test_jsonrpc.py`:** add `TestJSONRPCProtocol` — parse error, batch rejection,
  notification-no-response, error-uses-HTTP-200, and a success baseline (standalone master-free
  server on port 38391, raw POST for malformed/array bodies).

### Files
`quarkchain/cluster/jsonrpc.py`, `quarkchain/cluster/jsonrpc_server.py`,
`quarkchain/jsonrpc_client.py`, `quarkchain/cluster/tests/test_jsonrpc.py`

### DIFF
--- a/quarkchain/cluster/jsonrpc_server.py
+++ b/quarkchain/cluster/jsonrpc_server.py
@@ -91,12 +91,14 @@ class RpcMethods:
 
     async def dispatch(self, request_json: Dict[str, Any], context=None) -> Optional[Dict[str, Any]]:
         req_id = None
+        is_notification = False
 
         try:
             if not isinstance(request_json, dict):
                 raise InvalidRequest("Request must be object")
 
             req_id = request_json.get("id")
+            is_notification = "id" not in request_json
 
             if request_json.get("jsonrpc") != "2.0":
                 raise InvalidRequest("Invalid JSON-RPC version")
@@ -104,8 +106,6 @@ class RpcMethods:
             method = request_json.get("method")
             if not isinstance(method, str):
                 raise InvalidRequest("Method must be string")
-
-            is_notification = "id" not in request_json
 
             if method not in self._methods:
                 raise MethodNotFound()
@@ -139,13 +139,17 @@ class RpcMethods:
             }
 
         except JsonRpcError as e:
+            if is_notification:
+                return None
             return {
                 "jsonrpc": "2.0",
                 "error": e.to_dict(),
                 "id": req_id,
             }
         except Exception:
-            logger.exception("Internal JSON-RPC error for method %s", method)
+            if is_notification:
+                return None
+            logger.exception("Internal JSON-RPC error for method %s", locals().get("method", "<unknown>"))
             return {
                 "jsonrpc": "2.0",
                 "error": {
--- a/quarkchain/cluster/jsonrpc_server.py
+++ b/quarkchain/cluster/jsonrpc_server.py
@@ -92,8 +92,11 @@ class RpcMethods:
     async def dispatch(self, request_json: Dict[str, Any], context=None) -> Optional[Dict[str, Any]]:
         req_id = None
         is_notification = False
+        method = ""
 
         try:
+            if isinstance(request_json, list):
+                raise InvalidRequest("Batch requests not supported")
             if not isinstance(request_json, dict):
                 raise InvalidRequest("Request must be object")
 
@@ -147,7 +150,7 @@ class RpcMethods:
                 "id": req_id,
             }
         except Exception:
-            logger.exception("Internal JSON-RPC error for method %s", locals().get("method", "<unknown>"))
+            logger.exception("Internal JSON-RPC error for method %s", method)
             if is_notification:
                 return None
             return {
--- a/quarkchain/cluster/jsonrpc_server.py
+++ b/quarkchain/cluster/jsonrpc_server.py
@@ -147,9 +147,9 @@ class RpcMethods:
                 "id": req_id,
             }
         except Exception:
+            logger.exception("Internal JSON-RPC error for method %s", locals().get("method", "<unknown>"))
             if is_notification:
                 return None
-            logger.exception("Internal JSON-RPC error for method %s", locals().get("method", "<unknown>"))
             return {
                 "jsonrpc": "2.0",
                 "error": {
--- a/quarkchain/cluster/tests/test_jsonrpc.py
+++ b/quarkchain/cluster/tests/test_jsonrpc.py
@@ -2219,6 +2219,39 @@ class TestJSONRPCWebsocket(unittest.IsolatedAsyncioTestCase):
             response = json.loads(await websocket.recv())
             self.assertTrue(response["error"])
 
+    async def test_malformed_json_keeps_connection_alive(self):
+        # Malformed JSON must get a -32700 Parse error reply WITHOUT killing the
+        # connection (which would silently drop every subscription on it). After
+        # the error, a valid request on the same socket must still succeed.
+        id1 = Identity.create_random_identity()
+        acc1 = Address.create_from_identity(id1, full_shard_key=0)
+
+        async with ClusterContext(
+            1, acc1, small_coinbase=True
+        ) as clusters, jrpc_websocket_server_context(
+            clusters[0].slave_list[0], port=38601
+        ):
+            websocket = await get_websocket(port=38601)
+
+            # send invalid JSON -> expect a -32700 reply, not a closed socket
+            await websocket.send("{not valid json")
+            response = json.loads(await websocket.recv())
+            self.assertEqual(response["error"]["code"], -32700)
+            self.assertEqual(response["error"]["message"], "Parse error")
+            self.assertIsNone(response["id"])
+
+            # connection is still alive: a valid subscribe on the same socket works
+            request = {
+                "jsonrpc": "2.0",
+                "method": "subscribe",
+                "params": ["newHeads", "0x00000002"],
+                "id": 7,
+            }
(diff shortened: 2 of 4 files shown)


quarkchain/jsonrpc_client.py
```diff
@@ -28,6 +28,8 @@ def call(self, method, *params):
         resp.raise_for_status()
         data = resp.json()
 
+        if data.get("id") != payload["id"]:
```
qzhodl: Per spec, error responses where the server couldn't determine the id (e.g. -32700
Parse error, or rate-limit errors from public gateways) carry `"id": null`, so checking
the id before checking `"error"` masks the server's real error with a misleading
"id mismatch". Please check `"error"` first, and only validate the id on success responses.

quarkchain/cluster/jsonrpc_server.py
```diff
@@ -139,13 +139,17 @@ async def dispatch(self, request_json: Dict[str, Any], context=None) -> Optional
...
             return {
                 "jsonrpc": "2.0",
                 "error": e.to_dict(),
                 "id": req_id,
             }
         except Exception:
-            logger.exception("Internal JSON-RPC error for method %s", method)
+            if is_notification:
+                return None
+            logger.exception("Internal JSON-RPC error for method %s", locals().get("method", "<unknown>"))
```
qzhodl: Returning early here means internal errors in notification handlers are swallowed
without any log — a fire-and-forget request that crashes leaves no trace on the server.
Please move `logger.exception(...)` above the `if is_notification` check.

quarkchain/cluster/tests/test_jsonrpc.py
```diff
@@ -2216,3 +2218,116 @@ async def test_unsubscribe(self):
             await websocket.send(json.dumps(unsubscribe))
             response = json.loads(await websocket.recv())
             self.assertTrue(response["error"])
+
```
qzhodl, line 2221: The WS-side fix (malformed JSON now gets a -32700 reply instead of killing the
connection and all its subscriptions) is the most user-visible change here, but all five
new tests are HTTP-only. Could you add one WS test: send invalid JSON, assert the -32700
response, then send a valid request to verify the connection is still alive?

quarkchain/jsonrpc_client.py
```diff
@@ -28,8 +28,12 @@ def call(self, method, *params):
...
+            raise JsonRpcError({"code": -32600, "message": f"response id {data.get('id')!r} does not match request id {payload['id']!r}"})
         if "error" in data:
             raise JsonRpcError(data["error"])
+        if data.get("id") != payload["id"]:
```
qzhodl, line 33: This adds a second id check after the error check but leaves the original one
before it, so the masking bug is unchanged

qzhodl (approved):

blockchaindevsh (approved):
From Fable:

  Review: PR #988 — fix: JSON-RPC protocol compliance + protocol tests

  Verdict up front: Approve. All fixes are spec-correct, all prior reviewer comments are resolved at head 2afc2214, and CI is green.

  Overview

  This PR (4 commits, base upgrade/py313-baseline) hardens the home-grown JSON-RPC layer that replaced the old jsonrpcserver 3.5.4 dependency. It fixes three classes of spec violations and adds a protocol-level test suite:

  1. Malformed JSON — the HTTP server previously swallowed the json.loads failure (d = dict() fallback) and dispatched an empty dict, producing a misleading -32600 Invalid Request; the WS server raised InvalidParams, which escaped the async for
  loop and killed the connection, silently dropping every subscription on it (via the finally cleanup at jsonrpc.py:1527). Both now return a proper -32700 Parse error.
  2. JSON arrays / scalars over HTTP — the base code called d.get("method", ...) on whatever json.loads returned, so a batch array crashed with AttributeError → HTTP 500. Now guarded with isinstance and rejected cleanly with -32600.
  3. Notifications — dispatch() suppressed responses only on the success path; error paths still returned bodies. Per JSON-RPC 2.0, errors are now suppressed too.
  4. Client hardening — JsonRpcClient/AsyncJsonRpcClient now validate that the response id matches the request id.

  Correctness analysis

  Each change was verified against the base (7f695c9b) and head; the local working tree is byte-identical to the PR head for all four files.

  - jsonrpc.py:523-533 (HTTP __handle) — parse error returns {"jsonrpc":"2.0","error":{-32700},"id":null} with HTTP 200. Correct per spec (id must be null when it can't be determined). The isinstance(d, dict) guard at line 535 fixes the real
  AttributeError/500 on arrays; non-dict payloads flow into dispatch(), whose pre-existing check (jsonrpc_server.py:95-96) yields a clean -32600.
  - jsonrpc.py:1488-1496 (WS __handle) — sends -32700 and continues instead of raising, preserving the connection and its sub_ids state — the most user-visible fix. The isinstance guards on method/msg_id (lines 1497, 1503) prevent the same
  AttributeError for array/scalar frames over WS.
  - jsonrpc_server.py:94-101, 141-156 (dispatch) — is_notification is hoisted above validation and both except branches return None for notifications. Two subtleties, both fine: non-dict payloads raise InvalidRequest before is_notification is
  set, so the client still gets the -32600 body; and an object lacking "id" with a bad jsonrpc version gets no response — spec-ambiguous, but matches what jsonrpcserver did. logger.exception(...) is correctly placed before the notification
  early-return (line 152), so crashing fire-and-forget handlers still leave a server-side trace.
  - jsonrpc_client.py:31-34, 63-66 — the id check sits after the "error" check (the duplicate flagged in an earlier review comment was removed), so server errors carrying "id": null are no longer masked by a bogus "id mismatch". Verified in
  both sync and async clients.
  - No caller regressions: WS dispatch callers already handle None responses, and the HTTP handler's asyncio.shield wrapping is untouched.

  All four prior review comments from qzhodl (error-before-id ordering ×2, logging before notification return, missing WS test) are addressed at head.

  Code quality

  Minor nits, none blocking:
  - logger.exception("... %s", locals().get("method", "<unknown>")) works but is unidiomatic — pre-initializing method = "<unknown>" before the try would be cleaner.
  - The triple isinstance checks in WS __handle are repetitive; an early non-dict rejection would read better.
  - Arrays are valid JSON-RPC (batches), so a "Batch requests not supported" message would be friendlier than the generic "Request must be object". The no-batch deviation itself is intentional and documented.

  Test coverage

  Good. TestJSONRPCProtocol covers parse error, batch rejection (no 500), notification-no-response, error-with-HTTP-200, and a success baseline using a standalone master-free server (the func.__get__ binding trick in JSONRPCHttpServer.__init__
  was verified to work). The WS test test_malformed_json_keeps_connection_alive verifies both the -32700 reply and that a subsequent subscribe on the same socket succeeds — exactly the regression that matters. Acceptable gaps:
  array/notification frames over WS, explicit "id": null requests. Tests couldn't be run locally (no Python env with project deps), but CI runs them: all 4 checks green on head.

  Performance / security

  Neutral-to-positive: malformed input no longer produces unhandled exceptions, HTTP 500s, or torn-down WS subscription state — removes a trivial remote nuisance vector. Side effect: unparseable requests no longer increment self.counters, which
  is fine since counters only meaningfully tracked parsed methods.

  Optional follow-ups (non-blocking): pre-initialized method variable instead of locals().get, a "Batch requests not supported" message for arrays, and possibly HTTP 204 instead of an empty 200 body for notifications.

blockchaindevsh (approved):
From Fable:

  PR #988 incremental review (2afc2214 → 5548bac0)
  
  Changes: Only one new commit ("Improve JSON-RPC error messages and logging"), touching only jsonrpc_server.py — a direct response to the two non-blocking suggestions from the previous review:

  1. Pre-initialized method variable (jsonrpc_server.py:95): dispatch now initializes method = "" at the top, and the internal-error log uses it directly, replacing the previously flagged locals().get("method", "<unknown>") pattern. Verified
  that in that branch method can only be "" or an already-validated string — it will never log None or a stale value.
  2. Explicit batch-request rejection (jsonrpc_server.py:98-99): JSON arrays now return "Batch requests not supported" (error code still -32600, id: null), conforming to what JSON-RPC 2.0 prescribes for an invalid batch. The WS path goes
  through the same dispatch, so it benefits automatically.

  Status of prior feedback: Both suggestions have been addressed; the optional "HTTP 204 for notifications" suggestion was not adopted (existing tests pin the 200-with-empty-body behavior, which is acceptable); all of qzhodl's comments are
  resolved at head.

  CI: All 4 checks green.

  Verdict: Approve. Only two purely cosmetic points remain (the log shows an empty-string method when extraction fails; the batch test could also assert the new error message) — neither is blocking.

PR #989, opened 2026-06-11
fix: Make `Shard.add_block` cancellation-safe; stop cancelling in-flight handlers

ping-ke:
### Summary
Fixes a shard-sync hang. A per-connection `finally` cancelled in-flight message handlers, but
`Shard.add_block` is a multi-step commit with `await`s between steps and no failure cleanup. A
cancellation between awaits left `add_block_futures[hash]` permanently pending, so waiters and
`SyncTask` hung until the process was restarted.

### Related Issue

### Changes
- **`protocol.py`:** `active_and_loop_forever` no longer cancels `_handler_tasks` (restores
  run-to-completion). The fd-leak fix is kept (`writer.close()` on cancel, `reader.feed_eof()`), as
  is the `rpc_future_map` abort that unblocks any handler awaiting a dead connection's RPC.
- **`shard.py`:** wrap the commit protocol in `add_block` and `add_block_list_for_sync` with
  `try/except BaseException` → pop the future and `set_exception(RuntimeError(...))` on any
  interruption (including `CancelledError`), then re-raise. `RuntimeError` (not `CancelledError`) is
  used so a non-cancelled waiter isn't poisoned. Also fixes a pre-existing exception-path leak
  (broadcast to a dead slave). Sync now fails-and-retries on interrupted in-flight blocks instead of
  hanging.
- **`slave.py`:** remove the dead `SlaveServer.add_block_futures` (the live map is on `Shard`).
- **`tests/test_cluster.py`:** regression tests — a cancelled `add_block` cleans up its future, and
  a second waiter returns `False` instead of hanging forever.

### Files
`quarkchain/protocol.py`, `quarkchain/cluster/shard.py`, `quarkchain/cluster/slave.py`,
`quarkchain/cluster/tests/test_cluster.py`

### DIFF
--- a/quarkchain/cluster/shard.py
+++ b/quarkchain/cluster/shard.py
@@ -811,6 +811,10 @@ class Shard:
                 # forever.  Use RuntimeError to avoid propagating CancelledError
                 # into a non-cancelled waiter.
                 fut.set_exception(RuntimeError("add_block interrupted: {}".format(e)))
+                # Mark the exception as retrieved so asyncio does not log
+                # "Future exception was never retrieved" when no waiter consumes
+                # it (the common single-caller cancellation case).
+                fut.exception()
             raise
 
     def check_minor_block_by_header(self, header):
@@ -822,6 +826,27 @@ class Shard:
         if header.height == 0:
             return
         self.state.add_block(block, force=True, write_db=False, skip_if_too_old=False)
+
+    def _abort_uncommitted_add_block_futures(self, uncommitted_block_header_list, reason):
+        """Fail the add_block_futures we created for the given headers.
+
+        Waiters (and retries that would otherwise see BLOCK_COMMITTING and
+        gather() on these futures) must not hang forever.  Use RuntimeError, not
+        cancellation, so a non-cancelled waiter takes its `except Exception`
+        path instead of having CancelledError propagated into it.
+        """
+        for block_header in uncommitted_block_header_list:
+            block_hash = block_header.get_hash()
+            fut = self.add_block_futures.pop(block_hash, None)
+            if fut is not None and not fut.done():
+                fut.set_exception(
+                    RuntimeError(
+                        "add_block_list_for_sync interrupted: {}".format(reason)
+                    )
+                )
+                # Mark retrieved to avoid asyncio's "Future exception was never
+                # retrieved" log noise when no waiter consumes it.
+                fut.exception()
 
     async def add_block_list_for_sync(self, block_list):
         """ Add blocks in batch to reduce RPCs. Will NOT broadcast to peers.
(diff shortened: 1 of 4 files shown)


qzhodl (approved):

blockchaindevsh:
From Fable:

  Review: PR #989 — Make Shard.add_block cancellation-safe; stop cancelling in-flight handlers

  Overview

  A peer connection's finally block used to cancel in-flight message handler tasks. Shard.add_block is a multi-step commit (broadcast x-shard txs → notify master → commit → resolve future) with awaits between steps and no cleanup on
  interruption, so a cancellation mid-commit left add_block_futures[hash] pending forever — every later add_block/add_block_list_for_sync for that block hash awaited it indefinitely, and the shard stopped syncing until restart (issue #985).

  The fix is two-sided and the design is sound:
  - protocol.py restores run-to-completion for handler tasks, keeping the fd-leak fix and the rpc_future_map abort.
  - shard.py makes the commit sections cleanup-safe with except BaseException → pop the future, set_exception(RuntimeError(...)), re-raise. Waiters catch it and return False, so sync fails-and-retries instead of hanging.

  I verified the supporting details: write_rpc_request fails fast on a non-ACTIVE connection (protocol.py:106), and the finally in active_and_loop_forever sets state = CLOSED and aborts rpc_future_map in one synchronous block with no await in
  between, so an orphaned handler can't race a new RPC registration past the abort. Using RuntimeError instead of cancelling the future is correct: on Python 3.8+ except Exception would not catch CancelledError, so the waiter's except
  Exception: return False at shard.py:758-761 would otherwise be bypassed. The (False, None) return from add_block_list_for_sync is handled by its caller (slave.py:489 raises → error_code=1 → master retries). The removed
  SlaveServer.add_block_futures is genuinely dead — the only live map is on Shard. CI is green.

  Findings

  1. Residual future leak in add_block_list_for_sync — same hang class, different path (worth fixing in this PR)

  The try protecting the futures starts after the per-block loop, but futures are created inside the loop (shard.py:884). If state.add_block fails for block N (shard.py:876-878), the function does return False, None with futures for blocks
  1..N-1 already registered and nothing to clean them up. On the sync retry those blocks read as BLOCK_COMMITTING, land in existing_add_block_futures, and the gather hangs forever — exactly the bug this PR fixes. The same applies if
  get_root_block_header_by_hash(...) at shard.py:880 returns None (AttributeError escapes the function with futures leaked). Suggest popping any futures created so far in the early-return path, or restructuring so the cleanup except covers the
  loop too.

  2. "Future exception was never retrieved" log noise (minor)

  In the common cancellation case there is no concurrent waiter, so the RuntimeError set on the popped future is never consumed and asyncio logs a warning when it's garbage-collected — the new test_cancelled_add_block_resolves_future will
  itself emit this. Calling fut.exception() immediately after set_exception(...) suppresses the traceback logging without affecting real waiters (they still get the exception on await).

  3. Orphaned handlers now run detached (trade-off, acceptable — flagging for awareness)

  Not cancelling handler tasks means a handler blocked on something unrelated to the dead connection, with no timeout, now leaks instead of being reaped. With this PR the add_block futures always resolve and dead-connection RPCs fail fast, so
  the known blockers are covered, but any future handler that awaits an unbounded primitive will silently outlive its connection. The comment at protocol.py:199-204 documents the rationale well; just be aware the safety net is gone.

  4. Nits

  - AsyncMock is imported in test_cluster.py:3 but never used.
  - shard.py:915 is ~100 chars; would be reformatted if the project runs black (the rest of the file follows black's 88-char style).
  - Pre-existing, unchanged: add_block's docstring says "False on any error", but broadcast/master-RPC failures still raise (now with cleanup) rather than return False. Fine to leave, but the docstring could mention it.

  Tests

  The two regression tests are well-targeted: cancel-at-broadcast must clean up the future, and a second waiter must return False within a timeout instead of hanging. They use a real Shard with a mocked slave, hitting the actual commit-status
  state machine rather than mocking it. A third test covering finding 1 (batch sync where a mid-batch block fails validation after earlier futures were created) would round this out — and currently would fail, demonstrating the residual leak.

  Verdict

  Approve with one requested change: close the residual leak in add_block_list_for_sync's early-return path (finding 1), since it's the same hang the PR exists to eliminate and is cheap to fix here. Everything else is minor.

blockchaindevsh (approved):
From Fable:

  PR #989 incremental review (58b7c377 → 299b35c0)
  
  Changes: One new commit, touching only shard.py (+76/−46) and test_cluster.py (+1/−1), directly responding to the previous review:

  1. The only requested change from last time (the residual future leak in add_block_list_for_sync) — fixed correctly. The per-block loop has been moved inside the protecting try (shard.py:868-946), and the cleanup logic is extracted into a
  shared helper _abort_uncommitted_add_block_futures (shard.py:830-849). All three original leak paths are covered:
    - The return False, None early-exit on state.add_block validation failure — since return doesn't trigger except, the code explicitly calls the cleanup helper before returning (with a comment explaining exactly that);
    - The AttributeError from get_root_block_header_by_hash(...) returning None, and check() failures — both land in the outer except BaseException → cleanup → re-raise; the caller at slave.py:506-508 catches it and returns error_code=1, so
  sync retries normally.
    - The helper uses pop(..., None) + a not fut.done() check, so partially-committed batches (where some futures already had set_result and were deleted) are handled correctly; it still uses RuntimeError rather than CancelledError, so waiters'
  except Exception path is unaffected.
  2. The "Future exception was never retrieved" log noise — fixed: both set_exception sites are immediately followed by fut.exception() (shard.py:817, 849), so no destructor warning is produced when there is no waiter, and behavior is unchanged
  when waiters exist. This is the standard idiom.
  3. Nits partially addressed: the unused AsyncMock import was removed; the over-long line disappeared with the restructure. The docstring inconsistency is pre-existing and fine to leave.

  The subagent also diffed the moved loop body line by line — apart from re-indentation and the cleanup paths, it is semantically identical to the original — and checked the concurrency boundaries (concurrent callers for the same hash only join
  and never mutate the owner's future; gather(existing_add_block_futures) correctly stays outside the try, so it won't mistakenly abort futures belonging to others). No regressions found.

  Not addressed (non-blocking): The previously suggested third regression test (covering the validation-failure early-return path) was not added. Since that path now goes through the same shared helper exercised by the existing cancellation
  tests, it counts as optional.

  CI: All 4 checks green, and qzhodl has given an APPROVED review.

  Verdict: Approve.

PR #990, opened 2026-06-11
fix: Fix miner disable→restart lifecycle; bound test event-loop fd leak

ping-ke:
###  Summary

The miner's disable()/start() cycle was broken in two ways that left mining silently dead after a restart, and the test event-loop fixture leaked one fd per async/sync-test boundary.

### Related Issue

###  Changes

  - miner.py disable() (stale sentinel): when a subprocess is running, signal termination through the input_q sentinel only — do not also cancel the handler task. Cancelling left the subprocess's terminating None unconsumed in output_q, so the next start() read a stale None and exited immediately. The task is now cancelled only in the elif branch where no subprocess exists yet.
  - miner.py handle_mined_block() (dead-process restart): on the terminating None, join() the subprocess and reset self.process = None. Previously self.process was never cleared, so a subsequent start() posted work to the already-exited subprocess and mining silently never resumed. join() also reaps the zombie. Resetting here avoids racing the subprocess's async exit.
  - conftest.py: track and close the gap-filling loop once replaced by an IsolatedAsyncioTestCase loop, bounding the fd leak to one live loop.
  - tests/test_miner.py: regression tests for both disable() branches and the terminate→restart cycle.

### DIFF
--- a/quarkchain/cluster/miner.py
+++ b/quarkchain/cluster/miner.py
@@ -199,18 +199,29 @@ class Miner:
     def disable(self):
         """Stop the mining process if there is one"""
         if self.enabled and self.process:
-            # end the mining process
+            # Signal the subprocess to stop.  The None sentinel will be read by
+            # handle_mined_block which then exits cleanly.  Do NOT also cancel
+            # the task: cancelling leaves the sentinel in output_q unconsumed,
+            # so the next start() reads a stale None and exits immediately.
             self.input_q.put((None, {}))
-        self.enabled = False
-        if self._mining_task and not self._mining_task.done():
+        elif self._mining_task and not self._mining_task.done():
+            # No subprocess to signal; cancel the task directly.
             self._mining_task.cancel()
-            self._mining_task = None
+        self.enabled = False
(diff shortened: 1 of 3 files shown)


qzhodl (approved):

quarkchain/cluster/miner.py
```diff
@@ -199,18 +199,29 @@ def is_enabled(self):
...
         async def handle_mined_block():
             while True:
                 res = await self.output_q.coro_get()  # type: MiningResult
                 if not res:
+                    # The subprocess has put its terminating sentinel and is
+                    # exiting.  Reset the process handle so a subsequent start()
+                    # spawns a fresh subprocess instead of posting work to the
+                    # dead one (which would silently never mine).
+                    if self.process:
+                        self.process.join()
```
blockchaindevsh: Should call process.coro_join instead of process.join since it's in an asynchronous function?

blockchaindevsh (approved):

ping-ke:
Comments from Alan:
169a8428cde8123b9ec976d80a1e6d38

PR #997, opened 2026-08-26
add cache match test for python and pyethash

qizhou:
Pass test via `PYTHONPATH=. pytest ethereum/pow/tests/test_ethash.py::TestEthash::test_cache_pyethash_matches_python`

Also, `safe-sha3` seems to offer much higher performance than `pycryptodome`: `safe-sha3` took about 1.8s to generate the cache, while `pycryptodome` took 4.5s.


qzhodl (approved):
Test also passed on my side

PR #998, opened 2026-09-01
Dockerfile: support private repository builds with BuildKit secrets

ping-ke:
## Summary

- Support cloning private GitHub repositories during Docker builds.
- Pass the GitHub token through a BuildKit secret instead of a build argument.
- Keep unauthenticated cloning available for public repositories.
- Remove Git metadata and Git tooling after the image is built.
- Document public and private repository build commands.

## Usage

```bash
printf "GitHub token: "
read -rs GITHUB_TOKEN
echo
export GITHUB_TOKEN

docker build \
  --no-cache \
  --secret id=github_token,env=GITHUB_TOKEN \
  -f mainnet/singularity/Dockerfile \
  --build-arg GIT_REPOSITORY=https://github.com/QuarkChain/pyquarkchain-hot-fix.git \
  --build-arg GIT_TAG="<branch>" \
  -t "<docker image name>" \
  .

unset GITHUB_TOKEN
```

The GitHub token requires read access to the target repository. For a fine-grained PAT, `Contents: Read-only` is sufficient.

## Security

The token is mounted only for the clone command and is not stored in a Docker build argument or image layer.

## Verification

- Successfully cloned the private hot-fix repository using a BuildKit secret.
- `git diff --check` passes.

### DIFF
--- a/mainnet/singularity/Dockerfile
+++ b/mainnet/singularity/Dockerfile
@@ -1,15 +1,10 @@
 # syntax=docker/dockerfile:1
 
-FROM python:3.13-slim 
-LABEL maintainer="quarkchain"
-
-ARG GIT_REPOSITORY=https://github.com/QuarkChain/pyquarkchain.git
-ARG GIT_TAG=master
+FROM python:3.13-slim AS base
 
 RUN apt-get update && apt-get install -y \
     build-essential \
     gcc \
-    git \
     curl \
     libssl3 \
     libbz2-1.0 \
@@ -21,6 +16,14 @@ RUN apt-get update && apt-get install -y \
     zlib1g \
     vim \
   && rm -rf /var/lib/apt/lists/*
+
+FROM base AS builder
+
+ARG GIT_REPOSITORY=https://github.com/QuarkChain/pyquarkchain.git
+ARG GIT_TAG=master
+
+RUN apt-get update && apt-get install -y git \
+ && rm -rf /var/lib/apt/lists/*
 
 RUN pip install --upgrade pip
 
@@ -36,7 +39,7 @@ WORKDIR /code/pyquarkchain
 
 RUN echo "checkout to $GIT_TAG" && git checkout "$GIT_TAG"
 
-RUN pip install -r requirements.txt
+RUN pip install --prefix=/install -r requirements.txt
 
 RUN curl -fsSL https://s3-us-west-2.amazonaws.com/pyqkcmainnet/libqkchash.so -o /code/pyquarkchain/qkchash/libqkchash.so
 
@@ -44,6 +47,12 @@ RUN rm -rf /code/pyquarkchain/.git \
  && apt-get purge -y git \
  && apt-get autoremove -y
 
+FROM base AS runtime
+
+COPY --from=builder /install /usr/local
+COPY --from=builder /code/pyquarkchain /code/pyquarkchain
+
+WORKDIR /code/pyquarkchain
 
 EXPOSE 22 38291 38391 38491
 ENV PYTHONPATH=/code/pyquarkchain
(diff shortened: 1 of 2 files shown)


mainnet/singularity/Dockerfile
```diff
@@ -19,16 +24,26 @@ RUN apt-get update && apt-get install -y \
...
 
-ARG GIT_TAG
-RUN echo "checkout to $GIT_TAG" && git checkout $GIT_TAG
+RUN echo "checkout to $GIT_TAG" && git checkout "$GIT_TAG"
 
 RUN pip install -r requirements.txt
 
 RUN curl -fsSL https://s3-us-west-2.amazonaws.com/pyqkcmainnet/libqkchash.so -o /code/pyquarkchain/qkchash/libqkchash.so
 
+RUN rm -rf /code/pyquarkchain/.git \
```
qzhodl: `rm -rf` here does not remove the repository metadata from the image: the clone and checkout were committed by earlier `RUN` instructions, so their lower layers still contain the full `.git` pack and the initial working tree. Anyone who can pull the image can extract those layers and recover history or deleted files, which is especially risky for the new private-repository use case. Please perform clone + checkout + `.git` removal in a builder stage and copy only the cleaned working tree into the final image. A multi-stage build also keeps Git itself out of the final image layers.

mainnet/singularity/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Create a GitHub personal access token with read access to the target repository.
+For a fine-grained token, grant `Contents: Read-only` permission. If the
+organization uses SSO, authorize the token for that organization as well.
+
+Provide the token as a BuildKit secret so it is not stored in an image layer or
+passed as a Docker build argument:
+
+```bash
+read -rsp "GitHub token: " GITHUB_TOKEN
```
qzhodl: This prompt is Bash-specific. On zsh, which is the default shell on current macOS, `-p` means read from a coprocess and the documented command fails with `zsh:read:1: -p: no coprocess`. Please either state that this block must be run in Bash or use a prompt form that works in both Bash and zsh.

mainnet/singularity/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+Then build the image:
+
+```sh
+docker build \
+  --secret id=github_token,env=GITHUB_TOKEN \
+  -f mainnet/singularity/Dockerfile \
+  --build-arg GIT_REPOSITORY=https://github.com/QuarkChain/pyquarkchain-hot-fix.git \
+  --build-arg GIT_TAG="<branch>" \
```
qzhodl, line 42: One layer-cache caveat is worth documenting here: Docker does not re-run a `RUN` instruction merely because the remote branch has advanced. Rebuilding with the same `GIT_REPOSITORY` and `GIT_TAG` values can therefore reuse the cached clone and package an older commit; changing only `-t` does not invalidate that cache. Please recommend an immutable full commit SHA (preferred), or `--no-cache-filter builder` / `--no-cache` when the latest tip of a mutable branch is required. This behavior predates this PR, but the new branch-based example makes it relevant to users.

mainnet/singularity/README.md
```diff
@@ -0,0 +1,58 @@
...
+read -rs "GITHUB_TOKEN?GitHub token: "
+echo
+export GITHUB_TOKEN
+```
+
+Then build the image:
+
+```sh
+docker build \
+  --no-cache
```
syntrust: This build command is not runnable as written: the `--no-cache` line is missing the trailing `\`, so the shell terminates the command here.

qzhodl (approved):

syntrust (approved):

PR #999, opened 2026-09-03
P2: Harden RPC timeouts, cancellation cleanup, and concurrency limits

ping-ke:
## Summary

Prevent cancelled or stalled RPCs from leaking resources, bound minor-block sync time, and protect the HTTP JSON-RPC server from unbounded concurrent requests.

## Changes

- Clean up cancelled RPC futures and safely ignore late responses.
- Add timeouts for minor-block synchronization.
- Add concurrency and timeout protection to the HTTP JSON-RPC server.
- Add regression tests for cancellation, timeout, and overload handling.

## Testing

- `python -m pytest`
  - 417 items passed


qzhodl:
Why do we need this change? Is there a related tracking issue or root cause analysis we can refer to?

PR #1006, opened 2026-09-20
Add an in-place Python 3.13 upgrade path for legacy mainnet containers

ping-ke:
### Summary

Add an upgrade script and instructions for upgrading a `mainnet1.6.2` container whose node data is stored inside the container.

### Upgrade Flow

Validated with the following flow:

1. Start a `mainnet1.6.2` Docker node and let it run.
2. Stop the node process while keeping the container running.
3. Pull the latest code from `master`, run `upgrade_to_python313.sh`, and then run `source ~/.bashrc`.
4. Restart the node and confirm it runs normally with Python 3.13.


mainnet/singularity/README.md
```diff
@@ -48,3 +48,38 @@ unset GITHUB_TOKEN
...
+   ```bash
+   cd /code/pyquarkchain
+   git checkout master
+   git pull
+   ```
+
+3. Upgrade to Python 3.13 and activate the new environment:
+
+   ```bash
+   bash ./upgrade_to_python313.sh /code/pyquarkchain
```
qzhodl: May need to tell where to download this script

ping-ke: After we merge this PR, user just need pull latest code which contain this script.

